PR #58, opened 2026-03-03
feat: add recursive export and --overwrite flag

larkly:
## Description
Add recursive export and overwrite support to the `export` command. The `--recursive` flag exports a page and all its descendants into a nested directory tree. The `--overwrite` flag removes an existing export directory before re-exporting, preventing duplicate files and `(1)` suffixes on re-runs.

  ## Type of Change
  - [ ] Documentation update
  - [ ] Performance improvement
  - [ ] Code refactoring

  ## Testing

  ## Checklist

  ## Screenshots (if applicable)
  N/A

  ## Additional Context
  New CLI options added to `export`:

  | Flag | Description |
  |------|-------------|
  | `-r, --recursive` | Export page and all descendants into nested directories |
  | `--max-depth <n>` | Limit recursion depth (default: 10) |
  | `--exclude <patterns>` | Comma-separated title glob patterns to skip |
  | `--delay-ms <ms>` | Delay between page exports in ms (default: 100) |
  | `--dry-run` | Preview the page tree without writing files |
  | `--overwrite` | Remove existing export directory before re-exporting |

  Default behavior (without `--overwrite`) is unchanged — duplicate directories get `(1)` suffixes.

### DIFF
--- a/tests/export.test.js
+++ b/tests/export.test.js
@@ -8,7 +8,6 @@ const {
     isExportDirectory,
     uniquePathFor,
     exportRecursive,
-    sanitizeTitle,
   },
 } = require('../bin/confluence.js');
 
@@ -226,9 +225,7 @@ describe('exportRecursive', () => {
     const consoleErrSpy = jest.spyOn(console, 'error').mockImplementation(() => {});
 
     // Make readPage fail for the second child
-    let callCount = 0;
     client.readPage.mockImplementation(async (id) => {
-      callCount++;
       if (id === '3') throw new Error('network error');
       return '# content';
     });


pchuri:
Thanks for this PR! 🎉 The recursive export is a really solid addition — `--dry-run`, rate limiting via `--delay-ms`, and the `--exclude` patterns show great attention to real-world usage. Extracting `uniquePathFor` and `writeStream` into standalone functions is a nice cleanup too.

I've reviewed the changes and have a few suggestions:

---

### 1. `--overwrite` with `fs.rmSync` needs a safety guard (Medium)

`fs.rmSync(exportDir, { recursive: true, force: true })` will silently delete an entire directory tree. Since this CLI is heavily used by AI agents (no interactive prompts), if `--dest` points to a wrong location or `sanitizeTitle` produces an unexpected name, this could wipe unrelated files.

**Suggested approach — marker file:**

Write a hidden metadata file on export, and only allow `--overwrite` to delete directories that contain it:

```js
const EXPORT_MARKER = '.confluence-export.json';

function writeExportMarker(fs, path, exportDir, meta) {
  const markerPath = path.join(exportDir, EXPORT_MARKER);
  fs.writeFileSync(markerPath, JSON.stringify({
    exportedAt: new Date().toISOString(),
    pageId: meta.pageId,
    title: meta.title,
    tool: 'confluence-cli',
  }, null, 2));
}

function isExportDirectory(fs, path, dir) {
  return fs.existsSync(path.join(dir, EXPORT_MARKER));
}
```

Then guard the deletion:

```js
if (options.overwrite && fs.existsSync(rootExportDir)) {
  if (!isExportDirectory(fs, path, rootExportDir)) {
    throw new Error(
      `Refusing to overwrite "${rootExportDir}": not a confluence-cli export directory. ` +
      `Expected marker file ${EXPORT_MARKER} not found.`
    );
  }
  fs.rmSync(rootExportDir, { recursive: true, force: true });
}
```

This applies to both the single-page path (line 1249) and the recursive path (step 6). The marker also doubles as a debugging aid (`exportedAt`, `pageId`).

---

### 2. `require('path')` inside `uniquePathFor` (Low)

`uniquePathFor` receives `fs` as a parameter but calls `require('path')` internally. For consistency, either pass `path` as a parameter (like `fs`) or import it at the module level.

---

### 3. No test coverage for `exportRecursive` orchestration (Medium)

The underlying client methods (`getAllDescendantPages`, `buildPageTree`, etc.) are well-tested, but the `exportRecursive` function itself has no tests. Consider adding coverage for:
- `--dry-run` (no files written)
- `--overwrite` behavior (with and without marker)
- `--exclude` filtering in the recursive flow
- Partial failure handling (`failures` array)

---

### 4. `_referencedAttachments` shared state (Low)

`client._referencedAttachments` is overwritten on every `readPage()` call. This works for sequential processing but would break silently if the loop ever becomes parallel. A minor note for future-proofing — consider having `readPage()` return the set directly instead of relying on instance state.

---

### 5. Comment numbering (Nit)

Line 1477: `// 7. Summary` should be `// 8. Summary` — the previous step is also numbered 7.

---

Overall this is a great feature. Happy to approve once the `--overwrite` safety guard is in place. 👍

larkly:
Good suggestions. I've addressed your feedback in the latest commit

  ### 1. `--overwrite` safety guard (blocking)

  Added a `.confluence-export.json` marker file that gets written into the export directory on every export. Before `--overwrite` deletes a directory, it now
  checks for this marker and refuses to proceed if it's missing:

  Error: Refusing to overwrite "/path/to/dir" - it was not created by confluence-cli (missing .confluence-export.json).

  This applies to both the single-page and recursive export paths.

  ### 2. Fixed `require('path')` inside `uniquePathFor`

  Changed the signature from `(fs, dir, filename)` to `(fs, path, dir, filename)`, removing the inline `require('path')` call. Updated both call sites.

  ### 3. Added test coverage

  New `tests/export.test.js` with 11 tests covering:
  - `writeExportMarker` / `isExportDirectory` helpers
  - `uniquePathFor` conflict resolution
  - `exportRecursive`: dry-run, overwrite with/without marker, exclude filtering, partial failures, marker file creation

  ### 4. `_referencedAttachments` shared state

  Acknowledged as a known limitation -- no change needed since processing is sequential.

  ### 5. Fixed comment numbering

  `// 7. Summary` -> `// 8. Summary`

pchuri (approved):
LGTM! All feedback has been addressed cleanly. Marker file guard, path parameter consistency, test coverage, and comment numbering — all good. Thanks for the quick turnaround 👍

PR #72, opened 2026-03-10
fix: restrict config file permissions to owner only

tokuhirom:
### Description

Set config directory to 0o700 and config file to 0o600 to prevent other users from reading stored API tokens in multi-user environments.

### Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing

### Checklist

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -108,8 +108,11 @@ function readConfigFile() {
 function saveConfigFile(data) {
   if (!fs.existsSync(CONFIG_DIR)) {
     fs.mkdirSync(CONFIG_DIR, { recursive: true, mode: 0o700 });
+  } else {
+    fs.chmodSync(CONFIG_DIR, 0o700);
   }
   fs.writeFileSync(CONFIG_FILE, JSON.stringify(data, null, 2), { mode: 0o600 });
+  fs.chmodSync(CONFIG_FILE, 0o600);
 }
 
 // Helper function to validate CLI-provided options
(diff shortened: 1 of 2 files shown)


pchuri (commented):
Thanks for the contribution, @tokuhirom! Restricting config file permissions is an important security improvement.

I have one main concern:

### Existing files/directories are not remediated

`writeFileSync` with `{ mode: 0o600 }` and `mkdirSync` with `{ mode: 0o700 }` only apply permissions **when the file/directory is newly created**. If a user already has `config.json` at `0644` or `~/.confluence-cli/` at `0755` (the common upgrade path), those permissions remain unchanged after this patch.

**Suggested fix — add explicit `chmodSync` calls:**

```js
function saveConfigFile(data) {
  if (!fs.existsSync(CONFIG_DIR)) {
    fs.mkdirSync(CONFIG_DIR, { recursive: true, mode: 0o700 });
  } else {
    fs.chmodSync(CONFIG_DIR, 0o700);
  }
  fs.writeFileSync(CONFIG_FILE, JSON.stringify(data, null, 2), { mode: 0o600 });
  fs.chmodSync(CONFIG_FILE, 0o600);
}
```

This ensures that every save operation tightens permissions regardless of whether the file/directory existed before.

It would also be great to add test coverage for the `chmodSync` calls on existing files.

Other than that, the test refactoring (`captureConfigWrite` destructuring) looks clean. Nice work!

PR #75, opened 2026-03-12
Fix markdown-to-storage rendering on Confluence Cloud

eschoeller:
## Summary

Fixes several issues with `markdownToStorage()` / `htmlToConfluenceStorage()` when targeting Confluence Cloud instances:

- **Links not rendering**: Cloud no longer renders the `ac:link` + `ri:url` storage format for external URLs. This PR adds an `isCloud()` helper (checks if domain ends with `.atlassian.net`) and uses smart links (`<a href="..." data-card-appearance="inline">`) for Cloud instances while preserving `ac:link` format for Server/Data Center.

- **HTML entities in code blocks**: `markdown-it` HTML-encodes characters inside `<code>` blocks (`"` → `&quot;`, `&` → `&amp;`, etc.). These entities were being passed verbatim into CDATA sections, rendering as literal `&quot;` in Confluence code macros. Now decoded before CDATA insertion.

- **Trailing newline in code blocks**: `markdown-it` appends `\n` to code block content during HTML rendering, causing an extra blank line at the end of every Confluence code macro. Now trimmed.

- **Angle-bracket placeholders stripped**: A global `&lt;`/`&gt;`/`&amp;` decode at the end of `htmlToConfluenceStorage()` was converting `&lt;placeholder&gt;` back to `<placeholder>`, which Confluence then stripped as an unknown HTML tag. Removed the global decode — code blocks handle their own entity decoding separately.

## Test plan

- [x] Updated existing code block test to expect decoded entities in CDATA
- [x] Updated link test to expect Cloud smart link format
- [x] Added new test for Server/Data Center link format (`ac:link`) to ensure backward compatibility
- [x] All 140 tests pass

### DIFF
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -92,10 +92,6 @@ class ConfluenceClient {
     return d === 'api.atlassian.com' || this.apiPath?.includes('/ex/confluence/');
   }
 
-  isCloud() {
-    return this.domain && this.domain.trim().toLowerCase().endsWith('.atlassian.net');
-  }
-
   sanitizeApiPath(rawPath) {
     const fallback = '/rest/api';
     const value = (rawPath || '').trim();


pchuri (commented):
Thanks for the contribution, @eschoeller! This is a well-structured PR with clear explanations for each fix. Really appreciate you tackling multiple rendering issues together with proper test coverage.

### Overall

The changes look solid — the `isCloud()` helper, code block entity decoding, trailing newline trim, and the removal of the global entity decode are all sensible fixes. The Cloud vs Server/DC link branching preserves backward compatibility nicely, and the test additions cover both paths.

### One issue: entity decoding order

The current decoding chain has a potential double-decoding bug:

```javascript
const decodedCode = code.replace(/\n$/, '')
  .replace(/&quot;/g, '"')
  .replace(/&amp;/g, '&')   // &amp;lt; → &lt;
  .replace(/&lt;/g, '<')     // &lt; → <  (double-decoded!)
  .replace(/&gt;/g, '>');
```

If the original source contains a literal `&lt;` (i.e. the user wrote `&lt;` in their code), markdown-it would encode the `&` to produce `&amp;lt;`. With the current order, `&amp;` is decoded first → `&lt;`, then `&lt;` is decoded again → `<`. This silently corrupts the content.

Moving `&amp;` to the end fixes this:

```javascript
const decodedCode = code.replace(/\n$/, '')
  .replace(/&quot;/g, '"')
  .replace(/&lt;/g, '<')
  .replace(/&gt;/g, '>')
  .replace(/&amp;/g, '&');   // decode & last
```

### Minor note

The `isCloud()` heuristic (checking for `.atlassian.net`) is pragmatic and works for the vast majority of cases. Worth noting that custom-domain Cloud instances won't be detected, but that's an edge case we can address later with a config flag if needed.

Could you fix the decoding order? Everything else looks good to merge. 🙏

eschoeller:
Rebased on main (v1.27.3), no conflicts. All 145 tests passing. Ready to merge.

PR #84, opened 2026-03-26
fix(docs): correct search examples in SKILL.md to use --cql flag

sincerity711:
## Summary

- The search examples in SKILL.md showed raw CQL passed as a plain query string (e.g. `confluence search "type=page AND space=MYSPACE"`), but without the `--cql` flag this is treated as a literal text search (`text ~ "type=page AND space=MYSPACE"`), not a CQL query
- Added `--cql` option to the search command reference table
- Updated examples to use `--cql` with `siteSearch ~` which works more reliably on Confluence Server/DC

## Test plan

- [x] Verified `confluence search --cql 'siteSearch ~ "keyword" and space = "SPACE"'` returns correct results
- [x] Confirmed without `--cql`, the old example `confluence search "type=page AND space=MYSPACE"` does NOT execute as CQL

### DIFF
--- a/.claude/skills/confluence/SKILL.md
+++ b/.claude/skills/confluence/SKILL.md
@@ -197,16 +197,17 @@ confluence find "API Reference" --space MYSPACE
 Search pages using a keyword or CQL expression.
 
 ```sh
-confluence search <query> [--limit <number>]
+confluence search <query> [--limit <number>] [--cql]
 ```
 
 | Option | Default | Description |
 |---|---|---|
 | `--limit` | `10` | Maximum number of results |
+| `--cql` | false | Pass query as raw CQL instead of text search |
 
 ```sh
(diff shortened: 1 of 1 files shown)


pchuri (approved):
Thanks for the PR!

Good catch — the existing examples were passing CQL syntax without the `--cql` flag, which meant they were silently treated as literal text searches. The switch to `siteSearch ~` is also a nice improvement for Server/DC environments.

Minor suggestions (non-blocking):
- L701: The "Search and process results" section is meant to show general usage patterns, so `siteSearch ~ "release notes"` without the `space = "MYSPACE"` filter might be simpler and more universal here.
- The `--cql` default is listed as `false` in the options table — for a boolean flag, `-` or leaving it blank may be more consistent with the other options.

LGTM!

PR #87, opened 2026-04-01
security: fix CQL injection and path traversal vulnerabilities

yaowei1980:
- Add escapeCql() function to sanitize CQL query strings
  - Escapes backslashes, double quotes, wildcards (*, ?), and fuzzy match (~)
  - Applied to search() and findPageByTitle() methods

- Strengthen sanitizeTitle() to prevent path traversal
  - Remove '..' sequences
  - Remove leading dots (hidden files)
  - Replace filesystem reserved characters with underscore

- Add sanitizeFilename() function for safe filename handling
  - Used in uniquePathFor() for attachment downloads and exports

Fixes: CQL injection via search queries, path traversal via page titles and attachment filenames

All tests passing (159/159)

### Description
Brief description of what this PR does.

### Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing
- [ ] Tests pass locally with my changes
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] New and existing unit tests pass locally with my changes

### Checklist
- [ ] My code follows the style guidelines of this project
- [ ] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

Add screenshots to help explain your changes.

### Additional Context
Add any other context about the pull request here.

### DIFF
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -548,8 +548,9 @@ program
         fs.mkdirSync(destDir, { recursive: true });
 
         const uniquePathFor = (dir, filename) => {
-          const parsed = path.parse(filename);
-          let attempt = path.join(dir, filename);
+          const safeFilename = sanitizeFilename(filename);
+          const parsed = path.parse(safeFilename);
+          let attempt = path.join(dir, safeFilename);
           let counter = 1;
           while (fs.existsSync(attempt)) {
             const suffix = ` (${counter})`;
@@ -1351,9 +1352,21 @@ function isExportDirectory(fs, path, dir) {
   return fs.existsSync(path.join(dir, EXPORT_MARKER));
 }
 
+function sanitizeFilename(filename) {
+  if (!filename || typeof filename !== 'string') {
+    return 'unnamed';
+  }
+  return filename
+    .replace(/\.\./g, '')
+    .replace(/[\\/:*?"<>|]/g, '_')
+    .replace(/^\.+/, '')
+    .trim() || 'unnamed';
+}
+
 function uniquePathFor(fs, path, dir, filename) {
-  const parsed = path.parse(filename);
-  let attempt = path.join(dir, filename);
+  const safeFilename = sanitizeFilename(filename);
+  const parsed = path.parse(safeFilename);
+  let attempt = path.join(dir, safeFilename);
   let counter = 1;
   while (fs.existsSync(attempt)) {
     const suffix = ` (${counter})`;
@@ -1553,7 +1566,11 @@ function sanitizeTitle(value) {
   if (!value || typeof value !== 'string') {
     return fallback;
   }
-  const cleaned = value.replace(/[\\/:*?"<>|]/g, ' ').trim();
+  const cleaned = value
+    .replace(/\.\./g, '')
+    .replace(/[\\/:*?"<>|]/g, '_')
+    .replace(/^\.+/, '')
+    .trim();
   return cleaned || fallback;
 }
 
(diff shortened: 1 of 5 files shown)


pchuri (commented):
Thanks for the PR! Security hardening for CQL injection and path traversal is a valuable contribution.

A few comments on the changes:

### 1. `..` removal can be bypassed (Critical)

`sanitizeFilename()` and `sanitizeTitle()` both use:

```js
.replace(/\.\./g, '')
```

This is a single-pass replacement, so `....` becomes `..` after one pass — path traversal is still possible.

A safer approach would be using `path.basename()` to strip directory components entirely:

```js
function sanitizeFilename(filename) {
  if (!filename || typeof filename !== 'string') {
    return 'unnamed';
  }
  return path.basename(filename)
    .replace(/[\\/:*?"<>|]/g, '_')
    .replace(/^\.+/, '')
    .trim() || 'unnamed';
}
```

### 2. `sanitizeTitle` behavior change — `' '` → `'_'`

The existing code replaces reserved characters with a space, but this PR changes it to underscore. This could break compatibility with previously exported directories (e.g., re-exporting a page won't match the existing directory name).

Was this intentional? If so, it might be worth noting as a breaking change.

### 3. `escapeCql` — is `~` escape necessary?

`~` is a CQL operator (fuzzy match), but inside a quoted string literal it may not need escaping depending on the Confluence version. Could you verify this doesn't cause unexpected search behavior?

### 4. `package-lock.json` changes

The lock file includes unrelated dev dependency version bumps (`ajv`, `brace-expansion`, `flatted`, `minimatch`, `picomatch`). Consider splitting these into a separate commit or PR to keep the security fix focused.

pchuri:
As the maintainer, I've pushed a follow-up commit (875fc5e) to this branch addressing the review feedback. Thanks @yaowei1980 for kicking this off — the security direction is valuable, I just wanted to firm up the implementation before merging.

### What changed

**`sanitizeFilename` — robust traversal prevention**
The `.replace(/\.\./g, '')` approach is bypassable (e.g. `.a.` keeps one dot, certain interleavings survive). Switched to `path.basename()` after normalizing `\` → `/`, which structurally strips any directory component rather than depending on a regex pattern.

**`sanitizeTitle` — no breaking change, same security**
Dropped the `.replace(/\.\./g, '')` and reverted the space → underscore change. Existing exports stay compatible. Security still holds because every path separator is replaced with space, so the sanitized output can never contain a separator — `path.join(dir, sanitized)` cannot escape `dir`.

**`escapeCql` — injection-safe, semantics preserved**
Removed the `*`, `?`, `~` escapes. Those characters can't break out of a `"..."` CQL literal, so escaping them doesn't help injection-wise but does silently break wildcard searches users rely on. Kept the `\\` and `\"` escapes, which is what actually matters for injection.

Also added `\x00-\x1f` (control chars) to both sanitizers.

### Tests

Added 26 unit tests covering:
- CQL injection payloads attempting quote breakout
- Wildcard preservation (regression guard against over-escaping)
- `findPageByTitle` title + spaceKey escaping
- `sanitizeFilename` traversal (posix, windows, bypass vectors, edge cases)
- `sanitizeTitle` traversal and control-char stripping

185/185 tests pass, lint clean.

### Also

Reverted the unrelated dev-dependency bumps in `package-lock.json` to keep this PR focused on the security fix.

PTAL. If this looks good I'll merge.

PR #95, opened 2026-04-09
feat: add Claude Code marketplace plugin

frd1201:
## Summary

- Add marketplace configuration so the repo can be registered as a Claude Code plugin marketplace
- Move skill from `.claude/skills/` to `plugins/confluence/` following plugin conventions
- Add user-invocable `/confluence` slash command alongside automatic skill activation
- Update README with plugin installation instructions
- Update `install-skill` command to use new skill path

## Installation

```bash
/plugin marketplace add pchuri/confluence-cli
/plugin install confluence@pchuri-confluence-cli
```

Or directly:

```bash
claude --plugin-dir=/path/to/confluence-cli/plugins/confluence
```

## Test plan

- [x] `confluence install-skill` works with new path
- [x] All 171 existing tests pass
- [x] `npm pack` includes `plugins/` and `.claude-plugin/` in package
- [x] Plugin loads via `--plugin-dir` and `/confluence` command is available
- [x] Marketplace registration via `/plugin marketplace add` works


pchuri (commented):
Thanks for putting this together! The plugin marketplace structure looks clean and well-organized.

I have a few suggestions before merging:

### 1. Keep backward compatibility with the existing skill installation

The current PR **moves** `.claude/skills/` to `plugins/`, which breaks the existing `install-skill` workflow for users who don't use the plugin system. I'd recommend supporting **both** methods:

- Keep `plugins/confluence/skills/confluence/SKILL.md` as the single source of truth
- Add a `prepublish` (or `prepublishOnly`) script that copies it to `.claude/skills/confluence/SKILL.md`
- Keep the original `install-skill` source path in `bin/confluence.js` unchanged
- Include both `.claude/skills/` and `plugins/` in the `files` array in `package.json`

This way, plugin users get the new structure, and existing users can continue using `confluence install-skill` without any changes.

### 2. `.claude-plugin/` is likely excluded from the npm package

The root `.claude-plugin/` directory is a dotfile directory, which npm excludes by default. Since only `plugins/` is listed in the `files` array, `marketplace.json` won't be included in the published package. Please add `".claude-plugin/"` to the `files` array.

### 3. Version hardcoding in `plugin.json`

`plugins/confluence/.claude-plugin/plugin.json` has `"version": "1.28.0"` hardcoded. This will drift from `package.json` after the next release unless semantic-release is configured to update it. Consider either automating the sync or removing the version field if the schema allows it.

### 4. User-invocable `/confluence` command

The PR description mentions adding a user-invocable `/confluence` slash command, but the SKILL.md frontmatter doesn't include a `user-invocable` attribute. Could you verify that the plugin system picks this up correctly?

---

Minor: the `marketplace.json` version (`1.0.0`) vs `plugin.json` version (`1.28.0`) — is this intentional (marketplace schema version vs plugin version)?

Overall, the direction is great. Just needs a few tweaks to maintain backward compatibility and ensure everything is properly packaged. Looking forward to the update!

frd1201:
@pchuri Thanks for the feedback

Thanks for the detailed review, appreciated!

#### 1. Backward compatibility with `install-skill`

I considered this, but I would push back on making that change.

The `install-skill` command is the only documented interface for skill installation, and it already works correctly: it reads from `plugins/confluence/skills/confluence/SKILL.md` and writes to the user's `.claude/skills/` directory. I could not find any other code in the package that references `.claude/skills/` as a source path.

Adding a `prepublishOnly` copy script would create two copies of the same file in the npm package, which introduces maintenance risk through potential drift if the copy step is forgotten or fails silently. Since `.claude/skills/` inside the package was never intended as a public API, but only as an internal detail used by `install-skill`, I would prefer to keep a single source of truth under `plugins/` and avoid the extra copy mechanism.

Plugin users load the skill through the plugin system directly, and non-plugin users can continue using `confluence install-skill` as before.

#### 2. `.claude-plugin/` in `files` — Fixed ✅

Good catch.

I added `.claude-plugin/` to the `files` array and verified with `npm pack --dry-run` that `marketplace.json` is now included.

#### 3. Version in `plugin.json` — Fixed ✅

I removed the `version` field from `plugin.json`.

`semantic-release` only updates `package.json`, `npm-shrinkwrap.json`, and `CHANGELOG.md`, so keeping a hardcoded version in `plugin.json` would cause drift after each release. The plugin schema does not require this field, so removing it is the safer option.

#### 4. User-invocable `/confluence` command

The plugin system does not use a separate user-invocable frontmatter attribute.

A skill becomes user-invocable through the combination of `description`, `argument-hint`, and `allowed-tools` in its frontmatter. All three are already present in `SKILL.md`, so the `/confluence` slash command works through that mechanism. No separate `commands/` directory or additional attribute is required.

Also, `commands/` is a legacy format. For new plugins, `skills/` is the recommended approach.

#### Minor: `marketplace.json` vs `plugin.json` version

This is intentional. The version in `marketplace.json` (`1.0.0`) is the marketplace schema version, not the plugin version.

pchuri (approved):
Thanks for the thorough responses and the quick fixes! All feedback items have been addressed cleanly:

- `.claude-plugin/` added to `files` ✅
- `version` removed from `plugin.json` ✅
- Backward compatibility explanation makes sense — `install-skill` works as before ✅

LGTM, great work! 🚀

PR #99, opened 2026-04-11
fix: prepend /wiki context path in toAbsoluteUrl for Cloud attachment downloads

### LINKED ISSUE #98: attachments --download fails with 404 on Atlassian Cloud (missing version/modificationDate/cacheVersion query params)
**Describe the bug**

`confluence attachments <pageId> --download` always fails with `Error: Request failed with status code 404`, even though listing attachments on the same page (without `--download`) works correctly. Adding `--pattern` does not help. This happens on Atlassian Cloud.

I traced the root cause to the download URL construction: the Cloud `/download/attachments/<pageId>/<filename>` endpoint **requires** the query parameters `version`, `modificationDate`, and `cacheVersion`, which are returned inside each attachment's `_links.download` field from `/wiki/rest/api/content/<pageId>/child/attachment`. It looks like the CLI is either omitting or mis-constructing these parameters when building the download request, so the server rejects it with 404.

**To Reproduce**

Steps to reproduce (against Atlassian Cloud):

1. Pick any Confluence Cloud page that has at least one attachment. Example page ID: `123456789`.
2. List attachments — this works fine:
   ```sh
   confluence attachments 123456789
   # Found 1 attachment:
   # 1. example.png (ID: att987654321)
   #    Type: image/png • Size: 387 KB • Version: 1
   ```
3. Try to download the same attachment:
   ```sh
   confluence attachments 123456789 --download --dest /tmp/conf-test
   # Error: Request failed with status code 404
   ```
4. Narrowing with `--pattern` reproduces the same failure:
   ```sh
   confluence attachments 123456789 --pattern "*.png" --download --dest /tmp/conf-test
   # Error: Request failed with status code 404
   ```
5. Calling the REST API directly with the **same credentials** succeeds, proving the auth and permissions are fine — the issue is the URL the CLI sends:
   ```sh
   curl -sL -u "$EMAIL:$TOKEN" \
     "https://example.atlassian.net/wiki/rest/api/content/123456789/child/attachment" \
   | jq -r '.results[]._links.download'
   # /download/attachments/123456789/example.png?version=1&modificationDate=1700000000000&cacheVersion=1&api=v2

   curl -sL -u "$EMAIL:$TOKEN" \
     -o /tmp/conf-test/example.png \
     "https://example.atlassian.net/wiki/download/attachments/123456789/example.png?version=1&modificationDate=1700000000000&cacheVersion=1&api=v2"
   # → downloads a valid PNG
   ```
   Dropping any of `version` / `modificationDate` / `cacheVersion` from that URL causes the server to return 404, which matches the CLI's observed behavior.

**Expected behavior**

`confluence attachments <pageId> --download --dest <dir>` should download every attachment on the page to `<dir>`, using the full download URL (including `version`, `modificationDate`, and `cacheVersion` query parameters) from each attachment's `_links.download` field.

**Screenshots**

Not applicable — this is a CLI-only failure.

**Environment (please complete the following information):**

 - OS: macOS 26.4.1 (darwin 25.4.0, arm64)
 - Node.js version: v25.9.0
 - confluence-cli version: 1.29.0
 - Confluence version: Cloud (`*.atlassian.net`)
 - Auth: basic (email + API token), via `~/.confluence-cli/config.json` profile

**Additional context**

- Listing attachments via the CLI works — only `--download` is broken, so the bug is isolated to the download path / URL construction, not auth or page resolution.
- Reproduced on multiple Atlassian Cloud sites.
- Suggested fix: when downloading, use the `_links.download` value returned by `/rest/api/content/<pageId>/child/attachment` verbatim, prefixed with the instance base URL (`https://<domain>/wiki`), instead of rebuilding the path from the attachment ID or title alone. The query string returned by the API already contains all the parameters the Cloud download endpoint requires.
- Workaround I'm currently using: bypass the CLI and call the REST API + `curl -u "$EMAIL:$TOKEN"` directly with the full `_links.download` URL. That works reliably.

**Error logs**

```
$ confluence attachments 123456789 --download --dest /tmp/conf-test
Error: Request failed with status code 404

$ confluence attachments 123456789 --pattern "*.png" --download --dest /tmp/conf-test
Error: Request failed with status code 404
```

Only the message above is emitted — no stack trace is shown by the CLI. If there's a verbose/debug flag that would give a full stack, let me know and I'll re-run and attach it.

adamtan945:
## Summary

Fixes #98.

`confluence attachments <pageId> --download` always fails with `Error: Request failed with status code 404` on Atlassian Cloud, even though listing attachments on the same page works correctly.

### Root cause

The REST API returns each attachment's `_links.download` as a path **relative to the Confluence context root**, e.g.:

```
/download/attachments/123456789/example.png?version=1&modificationDate=1700000000000&cacheVersion=1&api=v2
```

On Atlassian Cloud the context root is `/wiki`, so the correct absolute URL is:

```
https://example.atlassian.net/wiki/download/attachments/123456789/example.png?version=...
```

But `downloadAttachment` calls `toAbsoluteUrl(attachment._links?.download)` (lib/confluence-client.js:864), which goes straight through `buildUrl` and produces:

```
https://example.atlassian.net/download/attachments/123456789/example.png?version=...
```

**Missing the `/wiki` prefix.** Cloud returns 404 for that URL.

The interesting part: the constructor already computes `this.webUrlPrefix = this.apiPath.startsWith('/wiki/') ? '/wiki' : ''` on line 38, and it's correctly prepended in four places when building web UI links (lines 416, 510, 1356, 1360). But it was never applied inside `toAbsoluteUrl`, so attachment downloads (and the `downloadLink` field on formatted attachments, line 2063) fell through the gap.

Query parameters (`version`, `modificationDate`, `cacheVersion`) are all preserved correctly — the original issue hypothesis about missing query params was wrong. The only thing missing is the context path.

### Fix

Prepend `webUrlPrefix` inside `toAbsoluteUrl` when the input is a relative path and doesn't already include the prefix. Server/DC behavior is unchanged because `webUrlPrefix` is an empty string on those instances.

```js
toAbsoluteUrl(pathOrUrl) {
  if (!pathOrUrl) return null;
  if (pathOrUrl.startsWith('http://') || pathOrUrl.startsWith('https://')) {
    return pathOrUrl;
  }

  const prefix = this.webUrlPrefix || '';
  const needsPrefix = prefix && !pathOrUrl.startsWith(`${prefix}/`);
  const withPrefix = needsPrefix ? `${prefix}${pathOrUrl}` : pathOrUrl;

  return this.buildUrl(withPrefix);
}
```

The double-prefix guard (`!pathOrUrl.startsWith(\`${prefix}/\`)`) protects against future API changes where `_links.download` might start returning paths that already include `/wiki`.

### Tests

Added three targeted tests to `tests/confluence-client.test.js`:

1. Cloud client prepends `/wiki` to a relative download path with full query string
2. Cloud client does not double-prepend when the path already starts with `/wiki/`
3. Server/DC client (no `webUrlPrefix`) leaves relative paths untouched — regression guard

Full suite: **174 passed / 174 total** (was 171 before).

## Test plan

- [x] `npx jest tests/confluence-client.test.js` — 103/103 passed (3 new)
- [x] `npx jest` — 174/174 passed
- [x] End-to-end: built locally with `npm link`, ran `confluence attachments <pageId> --download --dest /tmp/out` against a live Atlassian Cloud page, confirmed the attachment downloaded successfully (byte-identical to the file fetched via `curl` directly against the REST API's `_links.download`).
- [x] End-to-end before the fix: same command against the same page returned `Error: Request failed with status code 404`, confirming the reproduction.

### DIFF
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -2078,7 +2078,16 @@ class ConfluenceClient {
       return pathOrUrl;
     }
 
-    return this.buildUrl(pathOrUrl);
+    // Prepend context path (e.g. /wiki on Atlassian Cloud) when the API returns
+    // a path relative to the Confluence context root. Without this, URLs such as
+    // attachment download links (_links.download → "/download/attachments/...")
+    // get built as https://<domain>/download/... instead of
+    // https://<domain>/wiki/download/..., which returns 404 on Cloud.
+    const prefix = this.webUrlPrefix || '';
+    const needsPrefix = prefix && !pathOrUrl.startsWith(`${prefix}/`);
+    const withPrefix = needsPrefix ? `${prefix}${pathOrUrl}` : pathOrUrl;
+
+    return this.buildUrl(withPrefix);
   }
 
   parseNextStart(nextLink) {
--- a/tests/confluence-client.test.js
+++ b/tests/confluence-client.test.js
@@ -105,6 +105,39 @@ describe('ConfluenceClient', () => {
       });
       expect(httpClient.toAbsoluteUrl('https://cdn.example.com/file.pdf')).toBe('https://cdn.example.com/file.pdf');
     });
+
+    test('toAbsoluteUrl prepends /wiki context path on Atlassian Cloud', () => {
+      const cloudClient = new ConfluenceClient({
+        domain: 'test.atlassian.net',
+        token: 'token',
+        apiPath: '/wiki/rest/api'
+      });
+      // _links.download from the API is relative to the Confluence context root,
+      // so on Cloud (apiPath starts with /wiki/) we must prepend /wiki. Otherwise
+      // the resulting URL hits https://<domain>/download/... and returns 404.
+      expect(cloudClient.toAbsoluteUrl('/download/attachments/123/file.png?version=1&modificationDate=1700000000000&cacheVersion=1&api=v2'))
+        .toBe('https://test.atlassian.net/wiki/download/attachments/123/file.png?version=1&modificationDate=1700000000000&cacheVersion=1&api=v2');
+    });
+
+    test('toAbsoluteUrl does not double-prepend /wiki when path already starts with it', () => {
+      const cloudClient = new ConfluenceClient({
+        domain: 'test.atlassian.net',
+        token: 'token',
+        apiPath: '/wiki/rest/api'
+      });
+      expect(cloudClient.toAbsoluteUrl('/wiki/download/attachments/123/file.png'))
+        .toBe('https://test.atlassian.net/wiki/download/attachments/123/file.png');
+    });
+
+    test('toAbsoluteUrl leaves Server/DC paths untouched (no webUrlPrefix)', () => {
+      const serverClient = new ConfluenceClient({
+        domain: 'confluence.example.com',
+        token: 'token',
+        apiPath: '/rest/api'
+      });
+      expect(serverClient.toAbsoluteUrl('/download/attachments/123/file.png'))
+        .toBe('https://confluence.example.com/download/attachments/123/file.png');
+    });
   });
 
   describe('api path handling', () => {


adamtan945:
Closing as duplicate of #93, which was opened earlier and has the same core fix. I didn't notice #93 when I opened this — apologies for the noise. I've moved my root cause analysis, end-to-end verification, and the three regression tests over to #93 so the review can happen in one place.

PR #103, opened 2026-04-14
Follow-up improvements from mTLS PR review

anandghegde:
## Summary

Follow-up to #102, addressing the post-merge review suggestions from @pchuri.

- Warn when client key file permissions are more permissive than `0600` on Unix systems (private key security risk)
- Extract `mtlsCertQuestion()` helper to eliminate duplicated prompt definitions across `initConfig` interactive mode and `promptForMissingValues`
- Extract `validateMtlsProtocol()` helper to share the http+mTLS incompatibility check across CLI validation, env config, and profile config
- Update README mTLS examples to use platform-neutral paths (`~/.certs/`) instead of macOS-specific `/Users/you/.certs/`

## Validation

- `npm test -- --runInBand` — 178 tests passing
- `npm run lint` — clean

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,9 +144,9 @@ confluence --profile corp init \
   --domain "docs.example.com" \
   --api-path "/confluence/rest/api" \
   --auth-type "mtls" \
-  --tls-client-cert "/Users/you/.certs/client.pem" \
-  --tls-client-key "/Users/you/.certs/client.key" \
-  --tls-ca-cert "/Users/you/.certs/ca-chain.pem"
+  --tls-client-cert "~/.certs/client.pem" \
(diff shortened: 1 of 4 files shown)


pchuri (approved):
LGTM — clean follow-up that addresses all feedback from #102.

All four items landed correctly: private key permission check with a helpful `chmod 600` hint, `mtlsCertQuestion()` helper eliminating the triple duplication, `validateMtlsProtocol()` shared across all three config paths, and platform-neutral README paths. Net -39 lines is a nice bonus.

The bundled `follow-redirects` bump (GHSA-r4q5-vmmm-2653) is a welcome addition — good to get the auth header leak fix in alongside the mTLS work.

PR #107, opened 2026-04-16
Add forceCloud option for Cloud instances with custom domains

stevenhan2:
## Summary

Confluence Cloud instances with custom domains (e.g., `wiki.example.org` instead of `*.atlassian.net`) are misidentified as Server/Data Center by the `isCloud()` check. This causes links to render using the `<ac:link><ri:url>` storage format, which Confluence Cloud does not support — resulting in broken links on every uploaded page.

This PR adds a `forceCloud` config option that overrides the domain-based detection:

- **`lib/confluence-client.js`** — reads `forceCloud` from config and includes it in `isCloud()`
- **`lib/config.js`** — passes `forceCloud` through from both env-var and profile-based config paths
- **Environment variable**: `CONFLUENCE_FORCE_CLOUD=true`
- **Profile config**: `"forceCloud": true` in `~/.confluence-cli/config.json`

### The problem

| Detected as | Link format produced | Works on Cloud? |
| --- | --- | --- |
| Cloud (`*.atlassian.net`) | `<a href="url" data-card-appearance="inline">text</a>` | Yes |
| Server/DC (everything else) | `<ac:link><ri:url ri:value="url" />...</ac:link>` | **No** |

Custom domain Cloud instances fall into the "everything else" bucket.

### The fix

```js
// confluence-client.js constructor
this.forceCloud = !!config.forceCloud;

// isCloud() — 3 characters added
isCloud() {
  return this.isScopedToken()
    || (this.domain && this.domain.trim().toLowerCase().endsWith('.atlassian.net'))
    || this.forceCloud;
}
```

## Test plan

- [x] Added unit tests for `isCloud()` with `forceCloud: true` on custom domains
- [x] Added unit test verifying `forceCloud` produces Cloud link format (smart links) instead of `ac:link`
- [x] Added config tests for `CONFLUENCE_FORCE_CLOUD` env var
- [x] All 189 existing tests continue to pass
- [x] Updated README.md with custom domain documentation
- [x] Updated Claude Code skill documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,27 @@ export CONFLUENCE_API_TOKEN="your-scoped-token"
 
 `CONFLUENCE_API_PATH` defaults to `/wiki/rest/api` for Atlassian Cloud domains and `/rest/api` otherwise. Override it when your site lives under a custom reverse proxy or on-premises path. `CONFLUENCE_AUTH_TYPE` defaults to `basic` when an email is present and falls back to `bearer` otherwise. For `mtls`, set `CONFLUENCE_TLS_CLIENT_CERT` and `CONFLUENCE_TLS_CLIENT_KEY`; `CONFLUENCE_TLS_CA_CERT` is optional.
 
+**Custom domains on Confluence Cloud:**
+
+If your Confluence Cloud instance uses a custom domain (e.g., `wiki.example.org` instead of `*.atlassian.net`), the CLI may misidentify it as a Server/Data Center instance and produce broken link formats. Set `CONFLUENCE_FORCE_CLOUD=true` to override the automatic detection:
+
+```bash
+export CONFLUENCE_FORCE_CLOUD=true
+```
+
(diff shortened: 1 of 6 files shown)


pchuri (approved):
Thanks for the thorough PR — the problem statement, the comparison table, and the test coverage (unit + integration on the link format) made this easy to review. LGTM ✅

A couple of non-blocking follow-up thoughts:

1. **`apiPath` inference** — `inferApiPath()` in `lib/config.js` still only returns `/wiki/rest/api` for `*.atlassian.net`. Custom-domain Cloud users will likely need to set `apiPath: "/wiki/rest/api"` explicitly in addition to `forceCloud`. Worth mentioning in the README example, or having `forceCloud` also influence the inferred path for consistency.
2. **`confluence init` interactive flow** — doesn't prompt for `forceCloud`, so users can only discover it via docs / env var. Fine as a follow-up.

Nice work!

PR #115, opened 2026-04-23
feat: add json metadata and storage read support

madrover:
### Description
  Add machine-readable metadata and native storage read support to improve `confluence-cli` for automation and scripting use cases.

  This PR:
  - adds `confluence info --format json`
  - adds `confluence read <pageId> --format storage`
  - enriches `confluence children --format json` with stable metadata fields such as `id`, `title`, `type`, `status`, `spaceKey`, `parentId`, `version`, and `url`
  - adds tests and README/help updates
  - preserves existing human-readable output for current users

  ### Type of Change
  - [x] Documentation update
  - [ ] Performance improvement
  - [ ] Code refactoring

  ### Testing

  ### Checklist
  - [ ] I have commented my code, particularly in hard-to-understand areas

  Not applicable.

  ### Additional Context
  Local verification:
  - `npm run lint -- --quiet`
  - `npm test -- --runInBand`

  Notes:
  - existing default text output for `info`, `read`, and `children` remains unchanged
  - `info --format json` returns a single structured object
  - `read --format storage` returns the page body in Confluence storage representation through the normal read path
  - `children --format json` now returns structured metadata records for direct and recursive listings

### DIFF
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -1418,12 +1418,13 @@ class ConfluenceClient {
   /**
    * Get child pages of a given page
    */
-  async getChildPages(pageId, limit = 500) {
+  async getChildPages(pageId, limit = 500, options = {}) {
+    const includeAncestors = Boolean(options.includeAncestors);
     const response = await this.client.get(`/content/${pageId}/child/page`, {
       params: {
         limit: limit,
         // Fetch lightweight payload; content fetched on-demand when copying
-        expand: 'space,version,ancestors'
+        expand: includeAncestors ? 'space,version,ancestors' : 'space,version'
       }
     });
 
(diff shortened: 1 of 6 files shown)


pchuri:
Thanks for putting this together. The added docs and test coverage are appreciated, especially around the new JSON and storage-format flows.

lib/confluence-client.js
```diff
@@ -1429,17 +1423,13 @@ class ConfluenceClient {
       params: {
         limit: limit,
         // Fetch lightweight payload; content fetched on-demand when copying
-        expand: 'space,version'
+        expand: 'space,version,ancestors'
```
pchuri: Nice addition overall. One concern: `getChildPages()` is also used by `children --recursive`, recursive export, and `copy-tree`, so always requesting `ancestors` makes those paths fetch and retain the full ancestor chain even when they only need title/id/space. On large trees this can become a noticeable payload and memory regression. Could ancestor expansion be made opt-in for the JSON metadata path instead?

lib/confluence-client.js
```diff
@@ -1693,12 +1687,91 @@ class ConfluenceClient {
...
+    const fallbackUrl = (spaceKey && id)
+      ? `${this.webUrlPrefix}/spaces/${spaceKey}/pages/${id}`
+      : null;
+
+    return {
+      id,
+      title: raw?.title || '',
+      type: raw?.type || null,
+      status: raw?.status || null,
+      space,
```
pchuri, line 1761: `info --format json` is positioned as machine-readable output, but this currently forwards `raw.space` as-is. That leaks the Cloud/DC REST shape into the public JSON contract, so scripts may see different nested fields depending on deployment or API version. I think it would be safer to normalize `space` here, or omit it and rely on `spaceKey`.

pchuri (commented):
I left a couple of inline notes below.

lib/confluence-client.js
madrover: Good call. I changed getChildPages() so ancestor expansion is opt-in instead of always enabled. Recursive traversal now only requests ancestors for the JSON metadata path, so children --recursive, export, and copy-tree keep using the lighter payload by default.

lib/confluence-client.js
madrover: Agreed. I normalized space to a stable { key, name } shape in the machine-readable page metadata instead of passing through the raw REST payload. spaceKey is still present as the flat field for scripts that want the simplest contract.

madrover (commented):
Addressed both inline comments: ancestor fetching is now opt-in for the JSON metadata flow, and info --format json now returns a normalized space object. Tests and lint are passing locally.

pchuri (approved):

Issue #89, opened 2026-04-01
npm-shrinkwrap.json not published


jdiamond:
Hi, quick follow-up on #86. Thanks so much for addressing it so quickly. I'm still learning about how npm-shrinkwrap.json works so did some digging.

It looks like the published npm package still does not include `npm-shrinkwrap.json`.

I checked the current tarball with `npm pack confluence-cli --json`, and `npm-shrinkwrap.json` is not in the package contents. So `npm install -g confluence-cli` still seems to be resolving transitive dependencies from semver ranges.

My guess is that this may be related to the semantic-release config, since `.releaserc` currently only lists `package.json` and `CHANGELOG.md` as release assets.

I've never used semantic-release but I do see an example of this in their docs so might be right.

Thanks again for looking at this.

pchuri:
Thanks for the follow-up and for digging into this — you're right on both counts.

The `files` field in `package.json` only lists `bin/`, `lib/`, and `.claude/skills/`, which excludes `npm-shrinkwrap.json` from the published tarball. Additionally, `.releaserc` doesn't include `npm-shrinkwrap.json` in the `@semantic-release/git` assets, so the version field inside it won't be updated on release.

I'll push a fix for both shortly.

Issue #86, opened 2026-03-31
Harden npm installs by pinning axios and publishing shrinkwrap


jdiamond:
Hi, one supply-chain hardening note after the recent malicious axios publish on npm.

confluence-cli currently declares:

"axios": "^1.12.0"

That semver range would have allowed axios@1.14.1 during the incident window.

A repo package-lock.json helps maintainers locally, but it does not protect users installing the published CLI from npm. Per npm docs, package-lock.json is not published; npm-shrinkwrap.json is the publishable lockfile, and npm specifically recommends it for command-line tools intended as global installs.

I did not know this until today while checking my own install: your repo lockfile pins axios to 1.13.5, but my recent global install resolved 1.14.0. That seems consistent with published installs resolving from package.json ranges rather than the repo lockfile.

I’m not suggesting confluence-cli was compromised, only that the current dependency range leaves published installs exposed to dependency drift in a way that is avoidable for npm-distributed CLIs.

PR #116, opened 2026-04-24
fix: publish prod-only npm-shrinkwrap to avoid devDependency leakage

### LINKED ISSUE #110: Published `npm-shrinkwrap.json` causes 300+ extraneous dependencies for consumers
**Describe the bug**

`confluence-cli` publishes `npm-shrinkwrap.json` in its tarball. When consumers install it as a dependency, npm reads the shrinkwrap and installs ~310 nested packages (including `devDependencies` like `jest`, `eslint`, `@babel`, etc.) under `node_modules/confluence-cli/node_modules/`, all marked `extraneous`.

This also causes `npm install` and `npm install --package-lock-only` to produce different lockfiles, breaking tools like Renovate that rely on lockfile consistency.

**To Reproduce**

1. `mkdir test && cd test && npm init -y`
2. `npm install confluence-cli@1.30.1`
3. `grep -c extraneous package-lock.json` -> ~310
4. `curl -sL https://registry.npmjs.org/confluence-cli/-/confluence-cli-1.30.1.tgz | tar tz | grep shrinkwrap` -> confirms `npm-shrinkwrap.json` is in the tarball

**Expected behavior**

Consumers should only get the declared `dependencies`, not `devDependencies`. The published package should only lock production dependencies.

**Suggested fix**

Generate the shrinkwrap without `devDependencies` before publishing, e.g. `npm prune --omit=dev && npm shrinkwrap`, or use the [`--omit=dev` flag](https://docs.npmjs.com/cli/v10/commands/npm-shrinkwrap). This keeps the reproducibility benefit of shipping a shrinkwrap (appropriate for a CLI tool) without bundling ~310 extraneous `devDependencies` packages for consumers.

**Environment:**
- OS: macOS
- Node.js version: 24.14.1
- confluence-cli version: 1.30.1
- Confluence version: Cloud

**Additional context**

The shrinkwrap was introduced in response to #89 / #86.

pchuri:
### Description

Fixes #110. Consumers installing `confluence-cli` as a dependency were pulling in ~310 extraneous packages (`jest`, `eslint`, `@babel/*`, ...) under `node_modules/confluence-cli/node_modules/` because the published `npm-shrinkwrap.json` included `devDependency` entries.

This PR regenerates `npm-shrinkwrap.json` at publish time from a `package.json` with `devDependencies` stripped, so the published shrinkwrap only locks production dependencies — while still keeping the supply-chain hardening benefit from #86 / #89.

### Approach

- Repo keeps `package-lock.json` as the dev/CI lockfile (renamed from the previous `npm-shrinkwrap.json`, same content).
- `npm-shrinkwrap.json` is `.gitignore`d and only generated at publish time.
- New script `scripts/generate-prod-shrinkwrap.sh` resolves a fresh prod-only lockfile in a temp dir (`npm prune --omit=dev` does not actually remove `dev: true` entries from the lockfile, so we had to resolve from scratch).
- `.releaserc` runs the script via `@semantic-release/exec` in the `prepare` step, before `@semantic-release/npm` publishes. The shrinkwrap is no longer committed via `@semantic-release/git`.
- CI passes `@semantic-release/exec` as `extra_plugins` to `cycjimmy/semantic-release-action`.

### Type of Change

### Testing

Verified end-to-end locally against a packed tarball:

| | Before | After |
|---|---|---|
| `grep -c extraneous package-lock.json` in consumer project | **310** | **0** |
| Consumer `node_modules/confluence-cli/node_modules/` | `jest`, `eslint`, `@babel/*`, ... | (empty) |
| `npm ci` + `npm test` locally | pass | pass (253 tests) |
| Packed tarball contents | includes `npm-shrinkwrap.json` | includes `npm-shrinkwrap.json` (94 prod packages, 0 dev) |

- [x] Tests pass locally with my changes (253/253)
- [x] `npm ci` against the renamed `package-lock.json` resolves identically

### Checklist

### Additional Context

- Note to maintainers: the first release after this PR merges will be the first one that actually produces a clean tarball on npm. It would be worth smoke-testing a consumer install (`npm install confluence-cli@<new-version>` then `grep -c extraneous package-lock.json`) right after the release pipeline finishes to confirm the fix on the published artifact.


pchuri:
Follow-up on review feedback: fixed the registry re-resolution issue in [676ae72](https://github.com/pchuri/confluence-cli/pull/116/commits/676ae72).

**Problem confirmed**: The previous script wrote a `package.json` without `devDependencies` to a temp dir and ran `npm install --package-lock-only`, which re-resolved against the registry. Reproduced locally:
- `axios` 1.15.0 → 1.15.2
- `iconv-lite` 0.6.3 → 0.7.2

That means published tarballs could silently ship runtime versions that CI never tested — exactly the supply-chain drift the shrinkwrap was supposed to prevent (see #86).

**Fix**: The script now **filters** the checked-in `package-lock.json` — drops entries with `"dev": true` and clears `root.devDependencies` — instead of re-resolving. Runtime versions stay byte-identical to what CI validated.

**Pipeline reorder**: Also swapped `@semantic-release/npm` to run before `@semantic-release/exec` so the version bump lands in `package-lock.json` first, then the script generates the shrinkwrap at the new version. Added `package-lock.json` to the git assets so its bump is committed.

Verified end-to-end:
- `npm ci` + `npm test` + `npm run lint` pass on the dev lockfile
- Simulated full release flow (1.32.0 → 1.33.0 bump): generated shrinkwrap has version 1.33.0, axios 1.15.0 (matches checked-in), 0 dev entries
- Consumer install of the packed tarball: 0 extraneous, no nested `node_modules/confluence-cli/node_modules/`

PR #117, opened 2026-04-24
feat: add configurable linkStyle for storage-format link output

stevenhan2:
### Description

Some Confluence Cloud instances — particularly custom-domain Cloud setups — fail to render smart links (`<a data-card-appearance="inline">`) and surface "Cannot handle: DefaultLink" errors in their place. Plain `<a href>` tags render reliably everywhere.

This PR adds a `linkStyle` option with three values:

- `smart` — Cloud smart links with inline card appearance
- `plain` — simple `<a href>`, works on all Cloud instances
- `wiki`  — Server/DC `ac:link + ri:url` storage format

**No behaviour change unless explicitly opted in.** When `linkStyle` is unset, the converter falls back to the existing isCloud-based defaulting: `smart` for Cloud, `wiki` for Server/DC. Existing configurations — including `forceCloud: true` — produce byte-identical output to today.

Configurable via:

- `CONFLUENCE_LINK_STYLE` environment variable
- `linkStyle` key in a profile in `~/.confluence-cli/config.json`

Invalid values surface a warning at the config boundary and fall back to auto-detection; the converter itself is lenient to library consumers so a bad value can't break the pipeline.

Plumbed from `ConfluenceClient` into `MacroConverter` via a new constructor option. A `VALID_LINK_STYLES` constant is exported from `MacroConverter` for config-level validation to reuse (single source of truth, matching the existing `NAMED_ENTITIES` export pattern in `confluence-client.js`).

### Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [x] Documentation update (README + SKILL.md)

### Testing

- [x] Tests pass locally with my changes (267 passing, was 263)

New `tests/macro-converter.test.js` covers defaulting (Cloud/Server), explicit overrides per value, and each link-style output format. A pass-through test on `ConfluenceClient` confirms the option is plumbed correctly. `tests/config.test.js` adds env-var handling, invalid-value warning, and case-insensitive/trimmed normalization.

### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -249,6 +249,30 @@ Or add `"forceCloud": true` to your profile in `~/.confluence-cli/config.json`:
 }
 ```
 
+**Link rendering on Cloud (`linkStyle`):**
+
+Some Cloud instances — particularly custom-domain Cloud setups — fail to render smart links (`<a data-card-appearance="inline">`) and show "Cannot handle: DefaultLink" errors instead. If you hit this, set `linkStyle` to `plain` to emit simple `<a href>` tags, which render reliably everywhere:
+
+```bash
+export CONFLUENCE_LINK_STYLE=plain
+```
+
+Or per-profile:
+
+```json
+{
+  "profiles": {
+    "default": {
+      "domain": "wiki.example.org",
+      "forceCloud": true,
+      "linkStyle": "plain"
+    }
+  }
+}
+```
+
+Valid values: `smart` (Cloud smart links), `plain` (simple `<a href>`), `wiki` (Server/DC `ac:link`). When unset, the CLI picks `smart` for Cloud and `wiki` for Server/DC — existing behavior is unchanged.
+
 **Read-only mode** (recommended for AI agents):
 ```bash
 export CONFLUENCE_READ_ONLY=true
--- a/lib/config.js
+++ b/lib/config.js
@@ -17,6 +17,23 @@ const AUTH_CHOICES = [
 
 const AUTH_TYPES = ['basic', 'bearer', 'mtls', 'cookie'];
 
+const { VALID_LINK_STYLES } = require('./macro-converter');
+
+const normalizeLinkStyle = (rawValue, source) => {
+  if (rawValue === undefined || rawValue === null || rawValue === '') {
+    return undefined;
+  }
+  const value = String(rawValue).trim().toLowerCase();
+  if (VALID_LINK_STYLES.includes(value)) {
+    return value;
+  }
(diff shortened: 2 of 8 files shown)


pchuri (approved):
Thanks for the thoughtful submission, @stevenhan2 — this is a clean, well-scoped fix for a real-world Cloud rendering issue. LGTM. ✅

## What I liked

- **Byte-identical backward compatibility.** When `linkStyle` is unset, the converter falls back to the existing `isCloud`-based branching, so `forceCloud: true` users see no behavior change.
- **Correct precedence.** env (`CONFLUENCE_LINK_STYLE`) > profile > auto-detect, implemented cleanly with `??` at `lib/config.js:866-867`.
- **Defense-in-depth validation.** Invalid values warn at the config boundary, while the converter stays lenient so library consumers can't break the pipeline.
- **Good test coverage.** The new `tests/macro-converter.test.js` covers defaulting, explicit overrides, and all three output formats. The `ConfluenceClient` pass-through test nicely confirms the plumbing.
- **Single source of truth.** Exporting `VALID_LINK_STYLES` and reusing it in `lib/config.js` mirrors the existing `NAMED_ENTITIES` pattern — nice consistency.
- **Documentation.** README and SKILL.md updates are clear with concrete examples.

## Minor suggestions (non-blocking, feel free to defer)

1. **Add a profile-based `linkStyle` test.** `tests/config.test.js` only covers the env-var path. A test where `storedConfig.linkStyle = 'plain'` and `CONFLUENCE_LINK_STYLE` is unset would lock in the `??` fallback at `lib/config.js:866-867`. Bonus: a test asserting env overrides profile would nail down precedence.

2. **Persist `linkStyle` through `confluence init`.** Currently `saveConfig` at `lib/config.js:299-346` doesn't write `linkStyle`, so users have to hand-edit `~/.confluence-cli/config.json`. Worth a follow-up PR with a `--link-style` flag + interactive prompt.

3. **Warning side-effect location.** `normalizeLinkStyle(process.env.CONFLUENCE_LINK_STYLE, ...)` runs at the top of `getConfig()`, so an invalid env value warns even when the env-only branch isn't taken (profile path wins via `??`). Minor — could be deferred until the env branch is actually used.

4. **Pre-existing, out of scope:** the `<a href="(.*?)">(.*?)<\/a>` regex doesn't match `<a>` tags with additional attributes. Not introduced by this PR, but worth noting for any future work in this area.

Great work — approving. Thanks again for the detailed PR description and the incident context; it made the review much easier.

PR #118, opened 2026-04-24
feat: add TOC, ANCHOR, and same-page anchor link conversions

stevenhan2:
### Description

Extends `MacroConverter` with three paragraph-level marker conventions that mirror the existing `INFO` / `WARNING` / `NOTE` pattern:

- `**TOC**`            → Confluence Table of Contents macro (uses the macro's default heading levels — no hardcoded `minLevel` / `maxLevel`)
- `**ANCHOR: id**`     → Confluence anchor macro with the given id
- `> **QUOTE**`        → preserves a plain `<blockquote>`, opting out of the default conversion that wraps unmarked blockquotes in an info macro

Also adds same-page anchor link support: markdown links of the form `[text](#id)` convert to `ac:link` + `ac:anchor` so they render as in-page jumps in Confluence, matching how the anchor macro resolves. Runs before the general link conversion so the `#id` pattern is not consumed by the generic `<a href>` replacement.

No config or client changes — all contained in `lib/macro-converter.js` and a new `tests/macro-converter.test.js`.

### Type of Change

- [ ] Bug fix
- [ ] Breaking change
- [ ] Documentation update

### Testing

- [x] Tests pass locally with my changes (260 passing, was 256)

New `tests/macro-converter.test.js` covers each marker, the blockquote regression (unmarked blockquote still defaults to info), and the anchor-link ordering invariant on both Cloud and Server/DC.

### Checklist

Note on docs: the existing `INFO` / `WARNING` / `NOTE` markers are discoverable only from the code (not mentioned in README or SKILL.md), so this PR keeps parity rather than introducing a separate "marker conventions" section. Happy to add docs for all markers together in a follow-up if you'd prefer.

### DIFF
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -104,11 +104,6 @@ class MacroConverter {
         return `<ac:structured-macro ac:name="note">
           <ac:rich-text-body>${cleanContent}</ac:rich-text-body>
         </ac:structured-macro>`;
-      } else if (content.includes('<strong>QUOTE</strong>')) {
-        // QUOTE opts out of the default info-macro conversion and preserves
-        // a plain blockquote.
-        const cleanContent = content.replace(/<p><strong>QUOTE<\/strong><\/p>\s*/, '');
-        return `<blockquote>${cleanContent}</blockquote>`;
       } else {
         return `<ac:structured-macro ac:name="info">
           <ac:rich-text-body>${content}</ac:rich-text-body>
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -82,6 +82,12 @@ class MacroConverter {
 
     storage = storage.replace(/<code>(.*?)<\/code>/g, '<code>$1</code>');
 
+    // **TOC** paragraph → Confluence Table of Contents macro (uses macro defaults)
+    storage = storage.replace(
+      /<p><strong>TOC<\/strong><\/p>/g,
+      '<ac:structured-macro ac:name="toc" />'
+    );
+
     storage = storage.replace(/<blockquote>(.*?)<\/blockquote>/gs, (_, content) => {
       if (content.includes('<strong>INFO</strong>')) {
         const cleanContent = content.replace(/<p><strong>INFO<\/strong><\/p>\s*/, '');
@@ -98,6 +104,11 @@ class MacroConverter {
         return `<ac:structured-macro ac:name="note">
           <ac:rich-text-body>${cleanContent}</ac:rich-text-body>
         </ac:structured-macro>`;
+      } else if (content.includes('<strong>QUOTE</strong>')) {
+        // QUOTE opts out of the default info-macro conversion and preserves
+        // a plain blockquote.
+        const cleanContent = content.replace(/<p><strong>QUOTE<\/strong><\/p>\s*/, '');
+        return `<blockquote>${cleanContent}</blockquote>`;
       } else {
         return `<ac:structured-macro ac:name="info">
           <ac:rich-text-body>${content}</ac:rich-text-body>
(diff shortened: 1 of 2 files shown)


lib/macro-converter.js
```diff
@@ -77,6 +77,12 @@ class MacroConverter {
 
     storage = storage.replace(/<code>(.*?)<\/code>/g, '<code>$1</code>');
 
+    // **TOC** paragraph → Confluence Table of Contents macro (uses macro defaults)
+    storage = storage.replace(
+      /<p><strong>TOC<\/strong><\/p>/g,
+      '<ac:structured-macro ac:name="toc" />'
```
pchuri, line 88: **Round-trip asymmetry**: `storageToMarkdown` already strips TOC macros (lines 194–195), so a round trip silently drops the `**TOC**` marker. The existing `INFO` / `WARNING` / `NOTE` markers are bidirectional (line 223+), so there's a small parity gap here. Not a blocker — just worth calling out in the PR description, or considering an `[!toc]` reverse mapping in a follow-up so the convention stays symmetric.

lib/macro-converter.js
```diff
@@ -93,6 +99,11 @@ class MacroConverter {
         return `<ac:structured-macro ac:name="note">
           <ac:rich-text-body>${cleanContent}</ac:rich-text-body>
         </ac:structured-macro>`;
+      } else if (content.includes('<strong>QUOTE</strong>')) {
```
pchuri: The `content.includes('<strong>QUOTE</strong>')` heuristic matches the existing `INFO` / `WARNING` / `NOTE` pattern, so this is internally consistent — but it means any blockquote that *mentions* `**QUOTE**` anywhere in its body (not just as the first paragraph marker) gets classified as `QUOTE`. A stricter check like `content.startsWith('<p><strong>QUOTE</strong></p>')` would be safer. Same caveat applies to the three existing markers, so feel free to leave as-is for consistency or tighten all four together in a follow-up.

lib/macro-converter.js
```diff
@@ -107,6 +118,25 @@ class MacroConverter {
     storage = storage.replace(/<th>(.*?)<\/th>/g, '<th><p>$1</p></th>');
     storage = storage.replace(/<td>(.*?)<\/td>/g, '<td><p>$1</p></td>');
 
+    // **ANCHOR: id** paragraph → Confluence anchor macro
+    storage = storage.replace(
+      /<p><strong>ANCHOR: (.*?)<\/strong><\/p>/g,
+      '<ac:structured-macro ac:name="anchor"><ac:parameter ac:name="">$1</ac:parameter></ac:structured-macro>'
```
pchuri, line 124: **Round-trip asymmetry**: the catch-all at line 292 (`/<ac:structured-macro[^>]*>[\s\S]*?<\/ac:structured-macro>/g`) will silently strip this anchor macro on `storageToMarkdown`. If round-tripping is a goal, an `[!anchor: id]` reverse mapping would close the loop. Same as the `TOC` comment — non-blocking.

lib/macro-converter.js
```diff
@@ -107,6 +118,25 @@ class MacroConverter {
...
+    // before the general link conversion below so the #id pattern is not
+    // consumed by the generic <a href> replacement.
+    storage = storage.replace(/<a href="#(.*?)">(.*?)<\/a>/gs, (_, anchor, body) => {
+      const text = body
+        .replace(/&amp;/g, '&')
+        .replace(/&lt;/g, '<')
+        .replace(/&gt;/g, '>')
+        .replace(/&quot;/g, '"')
+        .replace(/&#39;/g, '\'');
+      return `<ac:link ac:anchor="${anchor}"><ac:plain-text-link-body><![CDATA[${text}]]></ac:plain-text-link-body></ac:link>`;
```
pchuri, line 138: Markdown formatting inside the anchor link text is lost — `[**bold jump**](#x)` becomes `<![CDATA[<strong>bold jump</strong>]]>`, which Confluence renders as literal text instead of bold. This matches the existing `wiki` linkStyle behavior on line 123, so it's not a regression; just flagging it in case you expect users to put rich text in anchor link labels.

pchuri (commented):
Thanks for the contribution! The marker convention extension is clean and well-tested — I really appreciate the explicit ordering invariant tests for both Cloud and Server/DC. Overall the implementation looks solid; I just have a few discussion points (left as inline comments) and one blocker before this can merge.

**Merge conflict**: this PR currently shows as `CONFLICTING` because #117 (linkStyle) merged after this branch was opened and added `tests/macro-converter.test.js` from scratch. Could you rebase onto `main` and merge your new `describe` blocks into the existing test file rather than overwriting it? The existing linkStyle tests need to stay.

The inline comments below are mostly observations about parity and edge cases — none are strict blockers, so feel free to address only what makes sense to you.

pchuri:
Rebased onto current `main` on the author's behalf — #117 (`linkStyle` config) had landed in the meantime and was conflicting with this branch.

Resolution:
- Restored `VALID_LINK_STYLES`, the `linkStyle` constructor option, and the `smart` / `wiki` / `plain` branching from #117 (which the original PR diff was inadvertently reverting).
- Placed the new anchor-link conversion **before** that branch so `[text](#id)` works under all three link styles — added a comment noting the `plain` mode case.
- All 338 tests pass; lint clean.

Force-pushed with `--force-with-lease`. Original commits and authorship preserved.

pchuri:
QUOTE marker dropped in `4bde196` to keep this PR scoped to TOC / ANCHOR / same-page anchor links — which are unambiguous additions.

The wider question (change the unmarked-blockquote default itself vs. keep adding an opt-out marker) is now tracked separately in #125.

CI green, mergeable. Ready when you are.

Issue #104, opened 2026-04-15
[BUG] Internal page links with custom display text are dropped during markdown conversion


tomerha:
**Describe the bug**
When converting a page to markdown with `confluence read <pageId> -f markdown`, internal page links that have a custom display text (via `<ac:link-body>`) are silently removed instead of being preserved as text. This causes table cells (and other elements) that contain only such links to appear empty in the output.

The `storageToMarkdown` method handles two `<ac:link>` patterns:

1. External URL links with `<ac:plain-text-link-body>` → converted to `[text](url)`
2. Internal page links without a link body → converted to `[Page Title]`

But internal page links with `<ac:link-body>` (the format Confluence uses when the author gives a link custom display text) aren't matched by either pattern, so they fall through to the catch-all removal on line 1428:

```js
markdown = markdown.replace(/<ac:link>[\s\S]*?<\/ac:link>/g, '');
```

Additionally, that catch-all only matches `<ac:link>` with no attributes. Links that carry attributes like `ac:anchor`, `ac:local-id`, or `ac:card-appearance` (e.g. `<ac:link ac:anchor="...">`) are not caught by this regex either, so they survive as raw HTML in the output.

**To Reproduce**
Any Confluence page with a table where cells contain internal links with custom display text. For example, storage format like:

```xml
<ac:link>
  <ri:page ri:content-title="Some Long Page Title" ri:version-at-save="28" />
  <ac:link-body>Short Name</ac:link-body>
</ac:link>
```

```sh
confluence read <pageId> -f markdown
```

The cell containing that link will be empty in the markdown table.

**Expected behavior**
The link's display text should be preserved. The above example should produce `Short Name` in the markdown output.

**Environment (please complete the following information):**
- confluence-cli version: 1.30.0
- Node.js version: v22.20.0
- OS: macOS

**Suggested fix**

Add a regex before the catch-all that extracts the display text from `<ac:link-body>`, and widen the catch-all to match `<ac:link>` tags with attributes:

```js
// Convert internal page links with custom link body text
markdown = markdown.replace(
  /<ac:link[^>]*>[\s\S]*?<ac:link-body>([\s\S]*?)<\/ac:link-body>[\s\S]*?<\/ac:link>/g,
  '$1'
);

// Remove any remaining ac:link tags that weren't matched
markdown = markdown.replace(/<ac:link[^>]*>[\s\S]*?<\/ac:link>/g, '');
```

pchuri:
[...]

PR #123, opened 2026-04-27
fix: preserve href on <a> tags during markdown conversion

### LINKED ISSUE #122: [BUG] Internal page links with data-card-appearance (smart links) lose URL during markdown conversion
**Describe the bug**

When converting a page to markdown with `confluence read <pageId> -f markdown`, internal page links rendered as **smart links / inline cards** (`<a href="..." data-card-appearance="inline">`) lose their URL. Only the display text survives, so the page cannot be round-tripped without destroying every cross-page link.

This is the sibling bug to #104, but affects a different storage element. #104 fixed `<ac:link>` with `<ac:link-body>`. The current bug affects ordinary `<a>` tags. Confluence Cloud's modern UI inserts cross-page references as smart links (`<a href="..." data-card-appearance="inline">`), so this is the default representation for new content — most pages on recent Confluence are affected.

**Storage format (input)**

```html
<ul>
  <li><p><a href="https://example.atlassian.net/wiki/spaces/SPACE/pages/135233540" data-card-appearance="inline">Linked Page Title</a></p></li>
</ul>
```

**Current markdown output (buggy)**

```
- Linked Page Title
```

The URL is silently dropped.

**Expected**

```
- [Linked Page Title](https://example.atlassian.net/wiki/spaces/SPACE/pages/135233540)
```

**To Reproduce**

1. Pick any modern Confluence Cloud page containing internal page references added via the inline-card UI (default for cross-doc links).
2. Run:

   ```sh
   confluence read <pageId> -f markdown
   ```

3. Observe that internal page links appear as plain text without URLs in the output.

**Why this matters**

After #104, `<ac:link>` round-trips work, but documentation suites that rely on cross-page navigation still break because new Confluence stores those references as smart-link `<a>` tags. An author who reads markdown, edits it, and pushes back loses every cross-doc link.

Note: the **opposite direction works correctly** — pushing markdown like `[text](url)` via `confluence update --format markdown` correctly produces `<a href="url" data-card-appearance="inline">text</a>` in storage format. The asymmetry is only on the read/export side.

**Environment**

- confluence-cli version: 1.33.1
- Node.js: v22.x
- OS: macOS

**Root cause**

`htmlToMarkdown` (`lib/html-to-markdown.js`) has **no handler for `<a href>` at all**. Anchor tags fall through to the generic attribute-strip regex on line 31 (`<(\w+)[^>]*>` → `<$1>`), losing every attribute including `href`, then to the catch-all on line 113 that replaces remaining unknown tags with a space. Only the display text survives.

**Suggested fix**

Add an explicit handler for `<a href>` ahead of the generic attribute-strip step. Allow attributes anywhere in the opening tag so smart-link metadata (`data-card-appearance`, `data-linked-resource-id`, …) does not block the match:

```js
markdown = markdown.replace(
  /<a\s+[^>]*?href="([^"]*)"[^>]*>([\s\S]*?)<\/a>/g,
  (_, href, text) => `[${text}](${href})`
);
```

Test cases to add:
- `<a>` with `data-card-appearance="inline"` (smart link)
- `<a>` with `data-linked-resource-id` and other smart-link attributes (varying attribute order)
- `<a>` inside table cells with smart-link attributes
- `<a>` inside list items
- Plain `<a href="url">text</a>` with no extra attributes

devlimits:
### Description

Internal page links rendered as smart links / inline cards (`<a href="..." data-card-appearance="inline">...</a>`) were silently losing their URL during `htmlToMarkdown` conversion. Only the display text survived, breaking round-trip workflows where authors read markdown, edit it, and push it back.

The `htmlToMarkdown` converter had no handler for `<a href>` at all. Anchor tags were stripped of their attributes by the generic attribute-strip regex on line 31, then dropped entirely by the catch-all on line 113, leaving only the display text.

This PR adds an explicit handler that converts `<a href="url">text</a>` to `[text](url)` before the generic attribute strip. The regex permits attributes anywhere in the opening tag so smart-link metadata (`data-card-appearance`, `data-linked-resource-id`, …) does not block the match.

Closes #122

### Type of Change

- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing

- [x] Tests pass locally with my changes (326/326)

### Checklist

### Additional Context

**Root cause:** `htmlToMarkdown` lacks an `<a href>` handler entirely. Smart links and inline cards — the default cross-page link format on modern Confluence Cloud — went through the generic attribute strip and the unknown-tag catch-all, so the URL was discarded.

**Asymmetry:** The push direction (`confluence update --format markdown`) correctly converts `[text](url)` back to `<a href="url" data-card-appearance="inline">text</a>` in storage format. The bug was only on the read/export side.

**Changes:**
- Added an `<a href>` regex handler in `htmlToMarkdown` ahead of the generic attribute-strip step
- Added 5 test cases covering plain anchors, smart-link `<a>` (`data-card-appearance="inline"`), anchors with `data-linked-resource-id` and varying attribute order, anchors inside table cells, and anchors inside list items

**Related:** Sibling fix to #104 / #105, which fixed the equivalent bug for `<ac:link>` / `<ac:link-body>`. After #104, modern Confluence pages still lost their cross-page navigation because Confluence Cloud now stores most internal references as smart-link `<a>` tags rather than `<ac:link>` macros.

### DIFF
--- a/lib/html-to-markdown.js
+++ b/lib/html-to-markdown.js
@@ -22,6 +22,14 @@ function htmlToMarkdown(html) {
 
   markdown = markdown.replace(/<time\s+datetime="([^"]+)"[^>]*(?:\/>|>\s*<\/time>)/g, '$1');
 
+  // Convert <a href="url">text</a> to [text](url) before generic attribute stripping.
+  // Allows attributes anywhere in the opening tag so smart links / inline cards
+  // (e.g. <a href="..." data-card-appearance="inline">) are preserved.
+  markdown = markdown.replace(
+    /<a\s+[^>]*?href="([^"]*)"[^>]*>([\s\S]*?)<\/a>/g,
+    (_, href, text) => `[${text}](${href})`
+  );
+
   markdown = markdown.replace(/<strong[^>]*>(.*?)<\/strong>/g, '**$1**');
 
   markdown = markdown.replace(/<em[^>]*>(.*?)<\/em>/g, '*$1*');
--- a/tests/html-to-markdown.test.js
+++ b/tests/html-to-markdown.test.js
@@ -21,6 +21,37 @@ describe('htmlToMarkdown', () => {
     });
   });
 
+  describe('anchor links', () => {
+    test('converts plain <a href> to markdown link', () => {
+      expect(htmlToMarkdown('<p><a href="https://example.com">Example</a></p>'))
+        .toBe('[Example](https://example.com)');
+    });
+
+    test('preserves URL on smart-link / inline-card anchors (data-card-appearance)', () => {
+      const html = '<p><a href="https://example.atlassian.net/wiki/spaces/X/pages/123" data-card-appearance="inline">Linked Page</a></p>';
+      expect(htmlToMarkdown(html))
+        .toBe('[Linked Page](https://example.atlassian.net/wiki/spaces/X/pages/123)');
+    });
+
+    test('preserves URL on anchors carrying smart-link metadata attributes', () => {
+      const html = '<p><a data-linked-resource-id="123" href="https://example.com/page" data-linked-resource-type="page">Title</a></p>';
+      expect(htmlToMarkdown(html))
+        .toBe('[Title](https://example.com/page)');
+    });
+
+    test('preserves anchor links inside table cells', () => {
+      const html = '<table><tr><th>Doc</th></tr><tr><td><a href="https://example.com/a" data-card-appearance="inline">A</a></td></tr></table>';
+      expect(htmlToMarkdown(html))
+        .toContain('[A](https://example.com/a)');
+    });
+
+    test('preserves anchor links inside list items', () => {
+      const html = '<ul><li><a href="https://example.com/x" data-card-appearance="inline">X</a></li></ul>';
+      expect(htmlToMarkdown(html))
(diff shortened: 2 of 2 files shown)


pchuri (approved):
LGTM — thanks for the careful fix and the thorough write-up!

A few things I appreciated while reviewing:

- The handler is placed **before** the generic attribute-strip on line 31, which is exactly where it has to go. Any later and `href` is already gone.
- Using `[^>]*?` + `[^>]*` around `href="..."` makes the match order-independent, which the `data-linked-resource-id`-first test case verifies nicely.
- Inner capture with `[\s\S]*?` (rather than `(.*?)` like the `<strong>`/`<em>` handlers) means multi-line anchors don't silently drop. Subtle but right.
- Interaction with the table/list cell cleanup is safe — by the time `<[^>]+>` runs over cells, anchors are already `[text](url)` and survive untouched.
- The 5 tests cover the realistic shapes: plain, smart-link `data-card-appearance`, attribute-order variation, inside `<td>`, inside `<li>`. All 37 tests pass locally.

This nicely closes the gap left after #104 — modern Confluence Cloud emits cross-page links as `<a data-card-appearance="inline">` rather than `<ac:link>`, so without this, round-tripping was still broken on most new pages.

Thanks again!

PR #130, opened 2026-04-27
feat(converter): add EXPAND marker for collapsible expand macro

stevenhan2:
## Summary

Adds a paragraph-level marker convention for the Confluence expand macro, mirroring the existing **TOC** and **ANCHOR** markers. Wrap content between `**EXPAND: title**` and `**EXPAND_END**` (each on its own paragraph) and it converts to a collapsible expand macro:

```markdown
**EXPAND: Show generated code**

\`\`\`js
const x = 1;
\`\`\`

**EXPAND_END**
```

The forward handler runs after code/blockquote/table conversion, so the captured body can already contain those macros. The reverse direction (`storageToMarkdown`) emits the same markers so the conversion round-trips — title-less expand macros (typically created in the Confluence UI without a title) still fall back to the existing `<details>/<summary>` output.

## Title sanitization — the non-obvious bit

Confluence's storage normalizer treats `<ac:parameter>` as text-only. I verified empirically against a Confluence Cloud instance:

| Title content                       | Confluence behavior                          |
|-------------------------------------|----------------------------------------------|
| `foo <em>bar</em> baz`              | silently truncated to `foo `                 |
| `foo <code>bar</code> baz`          | silently truncated to `foo `                 |
| `see <a href="..." ...>docs</a>`    | silently truncated to `see `                 |
| `<s>old</s> new`                    | **HTTP 500** — page creation rejected        |
| `A &amp; B`                         | works correctly (entities preserved)         |

So if a user writes a natural-feeling title like `**EXPAND: \`getUser()\` impl**`, markdown-it produces `<code>` inside the captured title, and Confluence then silently lops it off — the page would be created with title \"\" (just one word) without any warning. Strikethrough is even worse: a single `~~` makes `confluence create` fail with 500 at the API layer.

The forward handler strips inline HTML from the captured title with `title.replace(/<[^>]+>/g, '').trim()` before emitting `<ac:parameter>`. Entities (`&amp;`, `&lt;`) survive because the regex needs a literal `<`. Inline formatting in macro titles can't render anyway — silently dropping it matches the rendering reality and degrades gracefully.

## Test plan

- [x] `npx jest tests/macro-converter.test.js` — 35/35 pass
- [x] `npx jest` — full suite 361/361 pass
- [x] `npx eslint lib/macro-converter.js tests/macro-converter.test.js` — clean
- [x] Empirically verified against a live Confluence Cloud instance (silent truncation + 500-on-`<s>` findings above)

New tests cover: forward conversion, multiple EXPAND blocks per document (non-greedy regex correctness), title sanitization for `em`/`code`/`a`/`s`, entity preservation, titled round-trip, title-less fallback to `<details>`, and full markdown→storage→markdown round-trip.

## Notes

- README's \"Markdown Marker Conventions\" section gets a new entry alongside TOC and ANCHOR.
- The reverse handler is positioned before the existing generic `<details>` fallback so titled expand macros consume first.
- No behavior change for users whose Confluence pages contain title-less expand macros — they still convert to `<details>/<summary>` exactly as before.

### DIFF
--- a/README.md
+++ b/README.md
@@ -809,6 +809,24 @@ A paragraph containing only `**ANCHOR: my-section**` becomes a Confluence anchor
 **ANCHOR: my-section**
 ```
 
+### `**EXPAND: title** … **EXPAND_END**` — collapsible expand macro
+
+Wrap a block of content between `**EXPAND: title**` and `**EXPAND_END**` markers (each on its own paragraph) to render it as a Confluence expand macro with a collapsible body:
+
+````markdown
+**EXPAND: Show generated code**
+
+```js
+const x = 1;
+```
+
+**EXPAND_END**
+````
+
+The body may contain any content that is converted earlier in the pipeline (code blocks, tables, callout blockquotes). The reverse direction emits the same `**EXPAND: title**` / `**EXPAND_END**` markers so the conversion round-trips.
+
+Inline markdown inside the title (`*em*`, backtick code, links, `~~strike~~`) is stripped at capture time — Confluence's storage normalizer treats macro titles as plain text and will silently truncate or reject HTML in a `<ac:parameter>`. Title-less expand macros created in the Confluence UI still convert to `<details>/<summary>` blocks.
+
 ### `[text](#id)` — same-page anchor link
 
 A standard markdown link whose href starts with `#` becomes an `ac:link` with `ac:anchor`, rendering as an in-page jump in Confluence:
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -125,6 +125,20 @@ class MacroConverter {
       '<ac:structured-macro ac:name="anchor"><ac:parameter ac:name="">$1</ac:parameter></ac:structured-macro>'
     );
 
+    // **EXPAND: title** … **EXPAND_END** → Confluence expand macro. Runs
+    // after code/blockquote/table conversion so the body can contain those
+    // macros. Strips inline HTML from the title because Confluence's storage
+    // normalizer treats <ac:parameter> as text-only — it silently truncates
+    // at the first '<' and rejects <s> outright with HTTP 500. Entities
+    // (&amp;, &lt;) survive because the regex requires a literal '<'.
+    storage = storage.replace(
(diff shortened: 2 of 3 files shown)


pchuri (approved):
## Approving

Pattern mirrors the existing TOC/ANCHOR markers cleanly, and the title sanitization is empirically grounded — the documented Confluence storage normalizer quirks (silent truncation at the first `<`, HTTP 500 on `<s>`) make stripping inline HTML in the title the right call.

### Verified locally

- `npx jest tests/macro-converter.test.js` → **34/34 pass**
- `npx jest` → **927/927 pass**
- `npx eslint lib/macro-converter.js tests/macro-converter.test.js` → clean
- Live probes confirmed: entity round-trip (`A & B` survives via `htmlToMarkdown`'s `&amp;` decoder), body-with-code-block converts correctly, mixed titled + title-less expand macros in the same storage each route to the right handler.

### Nits (non-blocking)

1. **PR description test counts are off.** Macro-converter reports 34/34 (not 35/35), and the full suite is 927/927 (not 361/361 — that number looks stale). Worth fixing for accuracy.
2. **Commit message claims "body containing a code block" coverage but no such test is in the diff.** A body-with-code-block round-trip would also be a useful regression guard against pipeline ordering changes — worth adding, or drop the claim.

### Notes for the record (not blocking)

- The reverse regex assumes `<ac:parameter ac:name="title">` precedes `<ac:rich-text-body>`. If Confluence ever emits them reversed, the macro falls through to the existing title-less `<details>` handler — graceful degradation, worth a one-line comment so future readers know it's intentional.
- The body capture `[\s\S]*?` will mismatch when the body contains a nested `<ac:rich-text-body>` (e.g. an info macro inside an expand). Output happens to render correctly because subsequent regex passes absorb the dangling outer tags. This is the **same limitation** as the existing title-less expand handler, so this PR doesn't regress it — just flagging.
- A missing `**EXPAND_END**` leaves the open marker as a literal in the output without warning. Acceptable as graceful degradation.

LGTM — merge whenever the description nits are addressed.

PR #132, opened 2026-04-28
fix: strip callout markers and narrow detection to paragraph start

### LINKED ISSUE #131: INFO/WARNING/NOTE markers leak into macro body and trigger on prose mentions
### Summary

The `INFO` / `WARNING` / `NOTE` callout markers in `MacroConverter` produce two related issues that diverge from the documented behavior:

1. The marker text (`**INFO**` / `**WARNING**` / `**NOTE**`) leaks into the rendered macro body for the README-documented `> **INFO**\n> body` form.
2. Detection wraps any blockquote that merely mentions the marker keyword in prose (e.g. `> Use **INFO** at the start.`) in a callout macro.

#### 1. Marker text leak

```js
const MacroConverter = require('confluence-cli/lib/macro-converter');
const c = new MacroConverter({ isCloud: true });
console.log(c.markdownToStorage('> **INFO**\n> Heads up.'));
```

Actual output:
```xml
  <ac:rich-text-body>
    <p><strong>INFO</strong>
Heads up.</p>
  </ac:rich-text-body>
</ac:structured-macro>
```

Expected (per README and #5):
```xml
  <ac:rich-text-body>
    <p>Heads up.</p>
  </ac:rich-text-body>
</ac:structured-macro>
```

The same single-paragraph shape is what the `[!info]` preprocessor emits, so `[!info]\nbody` round-trip is also affected.

The cleanup regex on line 89 currently matches only the separated-paragraph form `<p><strong>INFO</strong></p>`, but markdown-it parses `> **INFO**\n> body` as a single paragraph `<p><strong>INFO</strong>\nbody</p>`, which never matches.

#### 2. False-positive macro wrapping

```js
console.log(c.markdownToStorage('> Use **INFO** at the start of a callout.'));
```

Actual output:
```xml
  <ac:rich-text-body>
    <p>Use <strong>INFO</strong> at the start of a callout.</p>
  </ac:rich-text-body>
</ac:structured-macro>
```

Expected: a plain `<blockquote>` (the marker keyword is just being mentioned, not used as a callout signal).

The detection step uses `content.includes('<strong>INFO</strong>')`, which matches any occurrence of the keyword anywhere in the blockquote.

### Impact

- All documented input paths produce visibly leaked marker text:
  - `> **INFO**\n> body` (#127 README form)
  - `> **INFO**\n>\n> body` works correctly only because the empty `>` line splits the paragraph
  - `[!info]\nbody` (#5 admonition form, via preprocessor expansion)
  - `> **WARNING**\n> body`, `> **NOTE**\n> body`, and the `[!warning]` / `[!note]` round-trip variants
- Authors quoting the marker keyword in prose get an unintended callout box

### Root cause

Both stem from the marker handling having different anchor semantics in detection vs. stripping:
- Detection: `content.includes('<strong>MARKER</strong>')` — too broad
- Stripping: `<p><strong>MARKER</strong></p>` — too narrow (only the separated-paragraph shape)

A single anchor — "the marker is the first thing inside the first paragraph, immediately followed by `</p>` or `\n`" — covers both correctly.

Happy to send a fix.

devlimits:
### Description

Closes #131.

The INFO/WARNING/NOTE callout markers in `MacroConverter` had two related issues that prevented the README-documented behavior from being reached.

#### 1. Marker text leaked into the macro body

The cleanup step only matched `<p><strong>MARKER</strong></p>` (separated paragraphs):

```js
content.replace(/<p><strong>INFO<\/strong><\/p>\s*/, '')
```

markdown-it parses the README form `> **INFO**\n> body` as a single paragraph `<p><strong>INFO</strong>\nbody</p>`, which never matched. The `[!info]` preprocessor emits the same single-paragraph shape, so both documented input formats produced macros with the marker text remaining in the body.

#5 originally introduced `[!info]` as an admonition that "renders as a Confluence info macro." #127 documented `> **INFO**\n> body` as the user-facing form. This change restores both forms to that behavior.

#### 2. Detection was too broad

`content.includes('<strong>INFO</strong>')` triggered an info macro for any blockquote containing the keyword anywhere in prose:

```markdown
> Use **INFO** at the start of a callout.
```

This was wrapped in an info macro and the marker text remained inside (since the strip regex correctly didn't match prose). Detection now uses the same anchor as stripping — the marker must sit at the start of the first paragraph, followed by `</p>` or `\n`.

#### Refactor

Three near-duplicate if/else branches (INFO/WARNING/NOTE) collapsed into a single `CALLOUT_MARKERS` lookup. The constant uses the canonical lowercase form (`['info', 'warning', 'note']`) matching `ac:name="info"` and the `[!info]` markdown shorthand. Uppercase conversion happens only at the marker-text matching site.

### Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Testing

- [x] Tests pass locally with my changes (363 passing, was 351)

12 new tests cover:
- Marker stripping for INFO/WARNING/NOTE × (single-paragraph form, separated form) — 6 tests
- `[!info]` / `[!warning]` / `[!note]` round-trip stripping — 3 tests
- Negative cases verifying that prose-level marker mentions stay plain blockquotes — 3 tests

### DIFF
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -2,6 +2,7 @@ const MarkdownIt = require('markdown-it');
 const { htmlToMarkdown } = require('./html-to-markdown');
 
 const VALID_LINK_STYLES = ['smart', 'plain', 'wiki'];
+const CALLOUT_MARKERS = ['info', 'warning', 'note'];
 
 class MacroConverter {
   constructor({ isCloud = false, webUrlPrefix = '', buildUrl = null, linkStyle = null } = {}) {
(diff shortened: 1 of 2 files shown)


pchuri (approved):
Thanks for the careful fix — this is exactly the right level. Locked detection and stripping to the same anchor (first paragraph start, followed by `</p>` or `\n`) which closes both the leak and the false-positive in one go. The case A / case B comment pinning down the markdown-it shapes is a great touch.

The `CALLOUT_MARKERS` refactor is a genuine improvement, not noise — three near-duplicate branches into one lookup, with the canonical lowercase form matching both `ac:name="info"` and the `[!info]` shorthand. Test coverage is solid: 12 new tests covering both forms × all three markers + the `[!info]` round-trip + three negative cases that guard against the false-positive regression.

Verified locally on the PR branch — full suite passes (929/929), and edge cases I poked at (marker-only blockquote, trailing punctuation, code-fence body, nested blockquotes, mid-prose mentions) all behave reasonably.

Nice work, LGTM!

PR #134, opened 2026-04-28
fix: scope `[!info]` admonition rewrite to block context

### LINKED ISSUE #133: Admonition preprocessor transforms [!info] regardless of markdown context
### Summary

The `[!info]` / `[!warning]` / `[!note]` admonition preprocessor in `MacroConverter.setupConfluenceMarkdownExtensions` runs a raw-text regex on `state.src` with no anchor, so any token that matches `[!info]` is rewritten — including inside fenced code blocks, inline code, heading text, and `>` blockquote prefixes. Three concrete failure modes follow.

#### 1. Heading inline code is shredded

```js
const MacroConverter = require('confluence-cli/lib/macro-converter');
const c = new MacroConverter({ isCloud: true });
console.log(c.markdownToStorage('## 5. `[!info]` round-trip marker'));
```

Actual output:
```html
<h2>5. `&gt; <strong>INFO</strong></h2>
<blockquote>
<p>` round-trip marker</p>
</blockquote>
```

The heading gets split mid-tag and the closing backtick ends up in a stray blockquote.

#### 2. Fenced code block contents are transformed

```js
console.log(c.markdownToStorage('```\n[!info]\nbody\n```'));
```

The `[!info]\nbody` inside the code block is rewritten to `> **INFO**\n> body`, breaking the fenced code rendering.

#### 3. `> [!info]` produces malformed XML (400 from Confluence)

```js
console.log(c.markdownToStorage('> [!info]\n> body'));
```

Actual output:
```html
  <ac:rich-text-body>
    <blockquote>
    <p><strong>INFO</strong>
body</p>
  </ac:rich-text-body>
</ac:structured-macro>
</blockquote>
```

The opening `<blockquote>` is inside the macro but its closing `</blockquote>` lands outside, producing unbalanced XML. Pushing this content via `confluence update` fails with HTTP 400.

### Root cause

`setupConfluenceMarkdownExtensions` operates on `state.src` *before* markdown-it tokenization. With no anchor, the regex matches `[!info]` anywhere in the source — code spans, fenced blocks, headings, and inside `>` blockquote prefixes. The `> [!info]` case additionally interacts with the storage handler's lazy `<blockquote>(.*?)</blockquote>` regex to produce the unbalanced output.

Happy to send a fix that scopes the rewrite to block context.

devlimits:
### Description

Closes #133.

The previous preprocessor ran a raw-text regex with no anchor, so any `[!info]` token in the source was rewritten — including inside fenced code blocks, inline code, headings, and `>` blockquote prefixes. The heading `## \`[!info]\` round-trip` ended up split mid-tag, and `> [!info]` GitHub-style alerts produced malformed XML that Confluence rejected with 400.

Two narrow changes contain the rewrite to its intended scope:

- **Stash fenced code blocks and inline code** with a Unicode private-use placeholder before the rewrite, then restore. Literal `[!info]` text inside code is no longer transformed.
- **Anchor the admonition pattern to a line start** (`(^|\n)\[!info\]`). Mid-paragraph mentions and `> [!info]` (preceded by a space, not a newline) no longer match, so the standard blockquote rule handles `> [!info]` as a plain quote — preventing the nested-blockquote XML mismatch that produced 400.

Also collapses three near-duplicate regex blocks into a `CALLOUT_MARKERS` loop and drops a no-op `- [ ]` task-list regex in the same function that captured and re-emitted the input unchanged.

### Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Testing

- [x] Tests pass locally with my changes (360 passing, was 351)

9 new tests cover the previously broken contexts:

- `[!info]` inside inline code (paragraph and heading)
- `[!info]` inside a fenced code block
- `[!info]` mentioned in heading text (no backticks)
- `[!info]` mid-paragraph
- `> [!info]` GitHub-style alerts deferring to plain blockquote

Plus 3 regression guards for the placeholder/restore step, ensuring user-authored numbers (in prose, headings, ordered list markers) survive the stash/restore round-trip and are not accidentally swapped for stash entries:

- digits in plain prose
- digits in a heading next to inline code
- ordered list markers coexisting with inline code

### Note on merge order with #132

This PR introduces \`const CALLOUT_MARKERS = ['info', 'warning', 'note'];\` at the top of \`lib/macro-converter.js\` and uses it in the new admonition loop. #132 introduces the same constant for the storage-side blockquote handler. Whichever PR lands second will hit a one-line declaration conflict — resolution is to keep a single declaration. The two PRs are otherwise independent and can land in either order.

### DIFF
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -3,6 +3,10 @@ const { htmlToMarkdown } = require('./html-to-markdown');
 
 const VALID_LINK_STYLES = ['smart', 'plain', 'wiki'];
 const CALLOUT_MARKERS = ['info', 'warning', 'note'];
+// U+E000 (Unicode Private Use Area) is used as the stash placeholder
+// delimiter. Declared as an explicit escape so the byte is visible in source
+// and survives editor / formatter / lint passes that strip invisible chars.
+const STASH_DELIM = '\uE000';
 
 class MacroConverter {
   constructor({ isCloud = false, webUrlPrefix = '', buildUrl = null, linkStyle = null } = {}) {
@@ -26,12 +30,11 @@ class MacroConverter {
     this.markdown.core.ruler.before('normalize', 'confluence_macros', (state) => {
       // Stash fenced code blocks and inline code so the admonition rewrite
       // below cannot transform `[!info]` tokens that the author intended as
-      // literal text inside code. The placeholder uses a Unicode private-use
-      // character that does not legitimately appear in markdown source.
+      // literal text inside code.
       const stash = [];
-      state.src = state.src.replace(/```[\s\S]*?```|`[^`\n]+`/g, (m) => {
+      state.src = state.src.replace(/```[\s\S]*?```|~~~[\s\S]*?~~~|`[^`\n]+`/g, (m) => {
         stash.push(m);
-        return `${stash.length - 1}`;
+        return `${STASH_DELIM}${stash.length - 1}${STASH_DELIM}`;
       });
 
       // Anchor `[!info]` to the start of a line (string start or after a
@@ -46,7 +49,11 @@ class MacroConverter {
         );
       }
 
-      state.src = state.src.replace(/(\d+)/g, (_, i) => stash[+i]);
+      // Fall back to the original match if the index is out of range so a
+      // literal U+E000<digits>U+E000 in user prose survives untouched instead
+      // of becoming the string "undefined".
+      const restoreRe = new RegExp(`${STASH_DELIM}(\\d+)${STASH_DELIM}`, 'g');
+      state.src = state.src.replace(restoreRe, (m, i) => stash[+i] ?? m);
     });
   }
 
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -2,6 +2,7 @@ const MarkdownIt = require('markdown-it');
 const { htmlToMarkdown } = require('./html-to-markdown');
 
 const VALID_LINK_STYLES = ['smart', 'plain', 'wiki'];
+const CALLOUT_MARKERS = ['info', 'warning', 'note'];
 
 class MacroConverter {
   constructor({ isCloud = false, webUrlPrefix = '', buildUrl = null, linkStyle = null } = {}) {
@@ -23,23 +24,29 @@ class MacroConverter {
     this.markdown.enable(['table', 'strikethrough', 'linkify']);
 
     this.markdown.core.ruler.before('normalize', 'confluence_macros', (state) => {
-      const src = state.src;
-
-      state.src = src.replace(/\[!info\]\s*([\s\S]*?)(?=\n\s*\n|\n\s*\[!|$)/g, (_, content) => {
-        return `> **INFO**\n> ${content.trim().replace(/\n/g, '\n> ')}`;
+      // Stash fenced code blocks and inline code so the admonition rewrite
+      // below cannot transform `[!info]` tokens that the author intended as
+      // literal text inside code. The placeholder uses a Unicode private-use
+      // character that does not legitimately appear in markdown source.
+      const stash = [];
+      state.src = state.src.replace(/```[\s\S]*?```|`[^`\n]+`/g, (m) => {
+        stash.push(m);
+        return `${stash.length - 1}`;
       });
 
-      state.src = state.src.replace(/\[!warning\]\s*([\s\S]*?)(?=\n\s*\n|\n\s*\[!|$)/g, (_, content) => {
-        return `> **WARNING**\n> ${content.trim().replace(/\n/g, '\n> ')}`;
-      });
+      // Anchor `[!info]` to the start of a line (string start or after a
+      // newline) so prose mid-paragraph, headings on the same line, and
+      // `> [!info]` GitHub-style alerts are left alone. The latter would
+      // otherwise expand to a nested blockquote that the storage handler's
+      // lazy regex cannot balance, producing malformed XML.
+      for (const m of CALLOUT_MARKERS) {
+        const re = new RegExp(`(^|\\n)\\[!${m}\\]\\s*([\\s\\S]*?)(?=\\n\\s*\\n|\\n\\s*\\[!|$)`, 'g');
+        state.src = state.src.replace(re, (_, pre, content) =>
+          `${pre}> **${m.toUpperCase()}**\n> ${content.trim().replace(/\n/g, '\n> ')}`
+        );
+      }
 
-      state.src = state.src.replace(/\[!note\]\s*([\s\S]*?)(?=\n\s*\n|\n\s*\[!|$)/g, (_, content) => {
-        return `> **NOTE**\n> ${content.trim().replace(/\n/g, '\n> ')}`;
-      });
-
-      state.src = state.src.replace(/^(\s*)- \[([ x])\] (.+)$/gm, (_, indent, checked, text) => {
-        return `${indent}- [${checked}] ${text}`;
-      });
+      state.src = state.src.replace(/(\d+)/g, (_, i) => stash[+i]);
     });
   }
 
--- a/tests/macro-converter.test.js
+++ b/tests/macro-converter.test.js
@@ -294,6 +294,25 @@ describe('MacroConverter markdownToStorage marker conventions', () => {
       expect(result).toContain('third');
       expect(result).not.toContain('undefined');
     });
+
+    test('[!info] inside a tilde-fenced code block is left intact', () => {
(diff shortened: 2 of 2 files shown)


devlimits:
Filed as #135 for separate tracking — round-trip drops multi-paragraph admonition body content outside the macro. Independent of this PR's scope.

lib/macro-converter.js
```diff
@@ -23,23 +24,29 @@ class MacroConverter {
...
-      });
-
-      state.src = state.src.replace(/\[!warning\]\s*([\s\S]*?)(?=\n\s*\n|\n\s*\[!|$)/g, (_, content) => {
-        return `> **WARNING**\n> ${content.trim().replace(/\n/g, '\n> ')}`;
+      // Stash fenced code blocks and inline code so the admonition rewrite
+      // below cannot transform `[!info]` tokens that the author intended as
+      // literal text inside code. The placeholder uses a Unicode private-use
+      // character that does not legitimately appear in markdown source.
+      const stash = [];
+      state.src = state.src.replace(/```[\s\S]*?```|`[^`\n]+`/g, (m) => {
```
pchuri: Tilde-fenced code blocks (`~~~`) aren't matched here, so `[!info]` inside one still gets rewritten by the admonition loop below. Reproduced on the PR branch:

```
input:  "~~~\n[!info]\nbody\n~~~"
output: <ac:structured-macro ac:name="code">…<![CDATA[> **INFO**
        > body
        > ~~~]]>…
```

Same class of bug as #133, just on the less common fence form. The PR description says "Stash fenced code blocks and inline code," so a reader would reasonably expect both fence forms covered. Suggested fix:

```js
state.src = state.src.replace(/```[\s\S]*?```|~~~[\s\S]*?~~~|`[^`\n]+`/g, ...)
```

Plus a regression test mirroring the existing fenced-code one.

lib/macro-converter.js
```diff
@@ -23,23 +24,29 @@ class MacroConverter {
...
+        stash.push(m);
+        return `${stash.length - 1}`;
```
pchuri: Heads up: this line and line 49 contain literal **U+E000** (Private Use Area) characters around the index, which most editors render zero-width and GitHub's diff viewer hides entirely. Reading the diff makes the placeholder look like bare digits:

```js
return `${stash.length - 1}`;
state.src = state.src.replace(/(\d+)/g, (_, i) => stash[+i]);
```

If that were the real code it would obviously corrupt any digit in user prose (`stash[+5]` → `undefined` → `"Version undefined released"`). The actual code is fine because the write and restore are both anchored to U+E000 delimiters — but the invisible bytes are a maintenance hazard:

- A reviewer reading the GitHub diff reaches the wrong conclusion about correctness.
- A future contributor's editor / auto-formatter / "trim invisible characters" lint rule could silently strip them, breaking the placeholder mechanism with no visible diff.

Could you write the delimiter as an explicit escape so it's visible? Something like:

```js
const STASH_DELIM = '\u{E000}';
state.src = state.src.replace(/```[\s\S]*?```|`[^`\n]+`/g, (m) => {
  stash.push(m);
  return `${STASH_DELIM}${stash.length - 1}${STASH_DELIM}`;
});
// ...
const restoreRe = new RegExp(`${STASH_DELIM}(\\d+)${STASH_DELIM}`, 'g');
state.src = state.src.replace(restoreRe, (_, i) => stash[+i]);
```

The comment on lines 27–30 should probably name U+E000 specifically too, so future readers don't have to re-derive it.

lib/macro-converter.js
```diff
@@ -23,23 +24,29 @@ class MacroConverter {
...
+        const re = new RegExp(`(^|\\n)\\[!${m}\\]\\s*([\\s\\S]*?)(?=\\n\\s*\\n|\\n\\s*\\[!|$)`, 'g');
+        state.src = state.src.replace(re, (_, pre, content) =>
+          `${pre}> **${m.toUpperCase()}**\n> ${content.trim().replace(/\n/g, '\n> ')}`
+        );
+      }
 
-      state.src = state.src.replace(/^(\s*)- \[([ x])\] (.+)$/gm, (_, indent, checked, text) => {
-        return `${indent}- [${checked}] ${text}`;
-      });
+      state.src = state.src.replace(/(\d+)/g, (_, i) => stash[+i]);
```
pchuri: Theoretical edge case (non-blocking): if a user's markdown literally contains `\u{E000}<digits>\u{E000}`, this replaces it with the string `"undefined"` (since `stash[that_index]` is out of bounds). Vanishingly unlikely — PUA characters don't naturally show up in markdown — but a one-character defense would harden it:

```js
state.src = state.src.replace(restoreRe, (_, i) => stash[+i] ?? _);
```

tests/macro-converter.test.js
```diff
@@ -151,6 +151,88 @@ describe('MacroConverter markdownToStorage marker conventions', () => {
     });
   });
 
+  describe('admonition preprocessor respects markdown context', () => {
```
pchuri, line 217: All 9 new tests in this block use `[!info]`. The `CALLOUT_MARKERS` loop in the source also handles `warning` and `note`, but only the `info` iteration is exercised here. (The earlier "marker conventions" tests cover `> **WARNING**` / `> **NOTE**`, but that's a different code path — the storage-side blockquote handler — so they don't pin the new loop.)

A one-line smoke test per marker would lock the loop in, e.g.:

```js
test('[!warning] at line start becomes a warning macro', () => {
  const result = converter.markdownToStorage('[!warning]\nheads up');
  expect(result).toContain('ac:name="warning"');
});
```

pchuri (commented):
Thanks for the careful write-up and the strong test set — the diagnosis (missing line-start anchor + raw rewrite over code spans) is exactly right, and collapsing the three marker blocks into `CALLOUT_MARKERS` is a nice cleanup. I checked out the branch and walked the new behavior through a few extra inputs. Inline notes below — two I'd ask you to address before merge, plus a couple of nice-to-haves.

Things I checked that are fine, for the record:

- **Indented (4-space) code blocks**: safe, because the new `(^|\n)\[!…\]` anchor disallows leading whitespace, so an indented `[!info]` no longer matches the admonition pattern at all. Nice side effect of the anchor change.
- **4-backtick fences**: the non-greedy stash regex doesn't capture both fences cleanly, but markdown-it tolerates the resulting fence-length mismatch and still produces a valid code block.
- **`> [!info]` deferring to plain blockquote**: the rationale (storage handler's lazy regex can't balance nested blockquotes) matches what I see in the storage-side code, and the test correctly pins the workaround.
- **Merge-conflict note with #132** is accurate and helpful.

lib/macro-converter.js
devlimits: Applied as suggested — added `~~~[\s\S]*?~~~` to the stash regex and a tilde-fence regression test.

lib/macro-converter.js
devlimits: Extracted into `const STASH_DELIM = '\uE000';` at the top of the file, with a comment explaining the explicit escape. Updated inline comments to name U+E000 explicitly.

lib/macro-converter.js
devlimits: Added the `?? m` fallback.

tests/macro-converter.test.js
devlimits: Added smoke tests for `[!warning]` and `[!note]`.

devlimits:
Thanks for the careful review - all four points addressed in the latest push.

pchuri (approved):
All four review points addressed cleanly — thanks for the fast and precise turnaround.

Verified on the updated branch:

- **Tilde fences**: `~~~\n[!info]\nbody\n~~~` now produces a clean code macro with the `[!info]` literal intact (previously corrupted into a `> **INFO**` blockquote with the closing `~~~` swallowed).
- **`STASH_DELIM` extraction**: the U+E000 byte is now visible in source as `'\uE000'` with an explanatory comment, and a full-file scan finds zero invisible PUA characters remaining.
- **Out-of-range defense**: `stash[+i] ?? m` correctly preserves a literal U+E000 + digits + U+E000 in user prose instead of replacing it with the string `"undefined"`. Confirmed with a manual reproduction of the prior corner case.
- **Marker coverage**: the new tilde-fence regression test plus `[!warning]` / `[!note]` smoke tests pin every iteration of the `CALLOUT_MARKERS` loop. Full suite is green (951 / 951).

One trivial nit (non-blocking): the constant is declared as `'\uE000'` (4-digit BMP escape) while the surrounding comments and commit message spell it `\u{E000}` (ES6 escape). Same code point either way — not worth a follow-up commit unless something else triggers one.

LGTM 🚢

PR #136, opened 2026-04-29
fix(converter): emit `> **MARKER**` form to close admonition round-trip

### LINKED ISSUE #135: [BUG/DISCUSS] Admonition body multi-paragraph round-trip drops content outside the macro
### Summary

When an admonition body contains multiple paragraphs (blank-line-separated), the second and subsequent paragraphs escape the macro on round-trip (download → re-upload). The bare `[!info]` form `storageToMarkdown` emits has no explicit body terminator, so a blank line inside the body is indistinguishable from the body ending.

Discovered while verifying #134.

#### Cycle 1 — author and push (works)

Author writes the admonition in the README-recommended `> **INFO**` form, with `>` on every line including the blank line between paragraphs:

```markdown
> **INFO**
> foo
>
> bar
```

`markdownToStorage` lands both paragraphs inside the macro as expected:

```html
  <ac:rich-text-body>
    <p>foo</p>
    <p>bar</p>
  </ac:rich-text-body>
</ac:structured-macro>
```

#### Cycle 2 — download (form changes)

`storageToMarkdown` emits the bare `[!info]` form, dropping the `>` prefix and using a blank line as the only paragraph separator:

```markdown
[!info]
foo

bar
```

#### Cycle 3 — re-push (breaks)

Pushing the downloaded markdown back, the preprocessor's lookahead `(?=\n\s*\n|...)` treats the blank line as the end of the admonition body. Only the first paragraph stays inside the macro; everything after the blank line escapes:

```html
  <ac:rich-text-body><p>foo</p></ac:rich-text-body>
</ac:structured-macro>
<p>bar</p>
```

### Root cause

The bare `[!info]` form has no explicit body terminator, so a blank line inside the body is indistinguishable from the body ending — the bare form has an inherent body-bounds ambiguity. The `>`-prefixed form authors naturally write *does* have an explicit terminator (where the `>` prefix stops), and `markdownToStorage` already handles it correctly, but `storageToMarkdown` doesn't emit that form.

### Possible directions

**Short-term:** have `storageToMarkdown` emit the README-recommended `> **INFO**` form for admonitions:

```markdown
> **INFO**
> foo
>
> bar
```

This form is already recognized by `markdownToStorage`, so round-trip closes. Body bounds are explicit via the `>` prefix and the blank-line ambiguity disappears.

**Longer-term:** GitHub Alerts–style support would be a nice direction too:

```markdown
> [!info]
> foo
>
> bar
```

This is familiar to GitHub / Obsidian users, has the same explicit-body-bounds property, and lets the bare `[!info]` form be deprecated entirely — eliminating the inherent ambiguity above. There are design decisions to settle (keyword mapping for NOTE/TIP/IMPORTANT/WARNING/CAUTION ↔ info/note/warning), so it's a bigger conversation.

Filing for awareness — happy to discuss further if you'd like to explore either direction.

pchuri:
## Description

Closes #135.

`storageToMarkdown` emitted the bare `[!info]` shorthand for callout macros, but the bare form has no explicit body terminator. A blank line inside a multi-paragraph body is therefore indistinguishable from the body ending, so re-uploading downloaded markdown silently dropped every paragraph after the first — content escaped the macro on round-trip.

### Reproduction (pre-fix)

Author writes the README-recommended form:

```markdown
> **INFO**
> foo
>
> bar
```

`markdownToStorage` correctly puts both paragraphs inside `<ac:rich-text-body>`.
`storageToMarkdown` then collapses to:

```markdown
[!info]
foo

bar
```

Re-uploading hits `lib/macro-converter.js`'s preprocessor `(?=\n\s*\n|...)` lookahead, which treats the blank line as the end of the body. `<p>bar</p>` lands outside the macro.

### Fix

Switch the storage→markdown output to the `> **MARKER**` blockquote form that README already documents as the canonical authored shape. The `>` prefix gives the body an explicit boundary, so `markdownToStorage` — which already handles this form correctly via the blockquote handler (PR #132) — preserves every paragraph on re-upload.

`markdownToStorage` still accepts the bare `[!marker]` form on input for backwards compatibility with already-downloaded files.

The three near-duplicate per-marker handlers (`info` / `warning` / `note`) collapse into a single `CALLOUT_MARKERS` loop, matching the pattern already used elsewhere in this file (PRs #132 and #134).

## Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Testing

- [x] Tests pass locally with my changes (391 passing, was 385)

6 new tests in a `storageToMarkdown callout round-trip` describe block:

- `info` / `warning` / `note` macros emit the `> **MARKER**` form (3 tests)
- Multi-paragraph body separates paragraphs with `>` lines (issue #135 direct repro)
- Full round-trip (`md → storage → md → storage`) preserves both paragraphs inside `<ac:rich-text-body>` and asserts no paragraph escapes the macro
- Single-paragraph round-trip preserves body without leaking marker text

The pre-existing `should convert Confluence macros to admonitions` test in `confluence-client.test.js` is updated to assert the new output form.

## Checklist

## Note on relationship to #132 / #134

This is the third bug in the same callout code path:

- #132 fixed marker text leaking into the rendered macro body (input side)
- #134 fixed the `[!info]` preprocessor running on tokens it shouldn't (input side, code/heading contexts)
- #135 / this PR fixes the output side emitting a form whose body bounds are ambiguous on re-upload

The author of #135 explicitly noted they discovered the bug while verifying #134.

### DIFF
--- a/README.md
+++ b/README.md
@@ -789,7 +789,7 @@ A blockquote whose first line is `**INFO**`, `**WARNING**`, or `**NOTE**` become
 > Side note for the reader.
 ```
 
-The reverse direction emits the equivalent shorthand (`[!info]` / `[!warning]` / `[!note]` followed by the body), which markdown→storage then re-expands.
+The reverse direction emits the same `> **INFO**` / `> **WARNING**` / `> **NOTE**` blockquote form, so multi-paragraph bodies round-trip cleanly. The bare `[!info]` / `[!warning]` / `[!note]` shorthand is still accepted on input for backwards compatibility.
 
 A blockquote without one of these markers stays a **plain blockquote** (`<blockquote>…</blockquote>`) — `> …` is treated as a quotation, not an alert. Use the markers above when you want a callout.
 
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -303,6 +303,12 @@ class MacroConverter {
     // boundary, so multi-paragraph callouts survive download → re-upload.
     // markdownToStorage still accepts the bare `[!marker]` form on input for
     // backwards compatibility with already-downloaded files.
+    //
+    // Wrap the output with leading + trailing `\n` to combine with the single
+    // `\n` that htmlToMarkdown emits around adjacent `<p>` tags. Without
+    // these, a following `<p>after</p>` lazy-continues into the blockquote
+    // body and lands inside the macro on re-upload (blockquote semantics —
+    // unlike code fences, blockquotes have no closing delimiter).
     for (const marker of CALLOUT_MARKERS) {
       const re = new RegExp(`<ac:structured-macro ac:name="${marker}"[^>]*>[\\s\\S]*?<ac:rich-text-body>([\\s\\S]*?)<\\/ac:rich-text-body>[\\s\\S]*?<\\/ac:structured-macro>`, 'g');
       markdown = markdown.replace(re, (_, content) => {
@@ -312,7 +318,8 @@ class MacroConverter {
           .map((line) => (line.length === 0 ? '>' : `> ${line}`))
           .join('\n');
         const header = `> **${marker.toUpperCase()}**`;
-        return body.length === 0 ? header : `${header}\n${quotedBody}`;
+        const inner = body.length === 0 ? header : `${header}\n${quotedBody}`;
+        return `\n${inner}\n`;
       });
     }
 
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -294,20 +294,27 @@ class MacroConverter {
       return `\n\`\`\`\n${code}\n\`\`\`\n`;
     });
 
-    markdown = markdown.replace(/<ac:structured-macro ac:name="info"[^>]*>[\s\S]*?<ac:rich-text-body>([\s\S]*?)<\/ac:rich-text-body>[\s\S]*?<\/ac:structured-macro>/g, (_, content) => {
-      const cleanContent = htmlToMarkdown(content);
-      return `[!info]\n${cleanContent}`;
-    });
-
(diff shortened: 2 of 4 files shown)


pchuri:
## Self-review follow-up: adjacency fix added

On re-review, the initial commit had a latent bug: the emitted `> **MARKER**` blockquote was not wrapped in separator newlines, so an adjacent following `<p>after</p>` would lazy-continue into the blockquote body on re-upload (markdown blockquotes have no closing delimiter — only a blank line ends them).

Reproduction (pre-followup):
```
storage:  <p>before</p><macro>foo,bar</macro><p>after</p>
                  ↓ download
markdown: before
          > **INFO**
          > foo
          >
          > bar
          after            ← lazy-continues
                  ↓ re-upload
storage:  <macro>foo, bar+after fused</macro>   ← `after` lands inside!
```

Follow-up commit ([c34b386](https://github.com/pchuri/confluence-cli/pull/136/commits/c34b386c4f8d1f874f7c77cfa00da9c419347b7c)) wraps the output with leading + trailing `\n` — same convention `code` and `mermaid` macros already use ([confluence-client.test.js:893-918](https://github.com/pchuri/confluence-cli/blob/main/tests/confluence-client.test.js#L893-L918)). The single `\n` combines with the single `\n` that `htmlToMarkdown` emits around adjacent `<p>` tags to produce blank-line separation.

Adds two regression guards in the new `callout round-trip` describe block:
- full round-trip with surrounding paragraphs preserves macro boundaries
- downloaded markdown has blank lines on both sides of the callout

Also updates the README's reverse-direction documentation to match the new output form. The bare `[!info]` shorthand is still accepted on input for backwards compatibility with already-downloaded files.

Test count: 391 → 393 passing.

PR #137, opened 2026-04-29
refactor(converter): replace regex-based storageToMarkdown with htmlparser2 walker

pchuri:
## Description

Replaces the regex-driven `storageToMarkdown` pipeline with an htmlparser2-backed DOM walker.

The previous implementation used 20+ non-greedy regex replacements over the raw storage XML. Nested macros (e.g. `expand` inside `panel`, `info` inside `expand`) caused the lazy `[\s\S]*?` matchers to land on the wrong closing tag — most recent fix commits in this area (#136, #134, #132) were workarounds for that structural fragility.

The new `lib/storage-walker.js` parses storage XML once via htmlparser2 in xmlMode, then walks the resulting DOM with a visitor pattern. Each Confluence macro and each HTML element has a dedicated handler. Nesting is handled by the tree itself rather than by regex ordering.

`MacroConverter.storageToMarkdown` shrinks from 165 lines to 11 lines.

## Type of Change
- [x] Code refactoring
- [x] Performance improvement (single parse pass replaces 20+ regex passes over the same string)

## What changed

| File | Change |
|---|---|
| `lib/storage-walker.js` (new) | DOM walker with handlers for `info`/`warning`/`note`, `expand`, `code`, `anchor`, `panel`, `mermaid-macro`, `include`, `shared-block`, `view-file`, `ac:link`, `ac:image`, `ac:task-list`, layouts, plus all standard HTML elements |
| `lib/macro-converter.js` | `storageToMarkdown` body delegated to the walker; removed the unused `htmlToMarkdown` import |
| `package.json` | Added `htmlparser2 ^9.1.0` (was already present transitively via `html-to-text`) |

`htmlToMarkdown` is preserved as a separate export — `lib/confluence-client.js` and the existing `html-to-markdown.test.js` continue to use it for plain-HTML input. Only the storage→markdown path was changed.

## Testing
- [x] Tests pass locally with my changes — **393/393 passing**
- [x] No new tests added — the refactor preserves existing contracts; the existing 103 macro-converter / html-to-markdown tests act as the regression net, plus round-trip coverage in `confluence-client.test.js`.

```
Test Suites: 11 passed, 11 total
Tests:       393 passed, 393 total
```

## Checklist
- [x] My code follows the style guidelines of this project (lint clean)

## Additional Context

Behavioral notes:
- Output formatting (line breaks, blank-line separators, blockquote `>` prefixes) is byte-identical to the previous regex pipeline for every existing test case.
- `<ac:link>` precedence order — `ac:anchor` → `ri:url` → `ac:link-body` → `ri:page` → drop — preserves the original regex ordering, so a `<ri:page>` paired with an explicit `<ac:link-body>` still renders the link-body text.
- xmlMode is required for proper CDATA parsing inside `<ac:plain-text-body>` / `<ac:plain-text-link-body>`. Default HTML mode mis-parses `<![CDATA[…]]>` as a directive node and loses the content.

### DIFF
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -1,5 +1,5 @@
 const MarkdownIt = require('markdown-it');
-const { htmlToMarkdown } = require('./html-to-markdown');
+const { StorageWalker } = require('./storage-walker');
 
 const VALID_LINK_STYLES = ['smart', 'plain', 'wiki'];
 const CALLOUT_MARKERS = ['info', 'warning', 'note'];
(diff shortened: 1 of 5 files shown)


pchuri:
Self-review pass — addressed the two blockers raised in code review:

**B1: Depth guard against pathological nesting**
- Added `StorageDepthExceededError` (typed) and a default cap of 256 levels in `StorageWalker`. Realistic Confluence layouts top out around 30, so the limit is generous but bounded — a single malformed payload now throws a recoverable error instead of crashing the Node process via V8 stack overflow.

**B2: `getTextContent` simplification**
- Removed the dead `cdata` branch that constructed a fake parent object to recurse. CDATA leaves are already handled correctly via the children path.

**New regression tests** (locking in capabilities the regex pipeline could not express):
- Triple-nested callouts (`info > warning > note`)
- Expand inside info
- Panel inside expand
- Reordered macro attributes (`ac:macro-id` before `ac:name`)
- Depth guard: 1000-level chain throws `StorageDepthExceededError`; 50-level chain succeeds

**Test status**: 399/399 passing (+6 new), lint clean.

**Deferred to follow-up issues** (all pre-existing on `main`, not introduced by this PR):
- S1: `cleanup()` strips leading whitespace inside fenced code blocks
- S3: `<s>` / `<del>` strikethrough lost on round-trip
- S4: External `<ri:url>` images dropped
- S5: Unescaped `]` `(` in `ri:content-title` can break markdown link syntax
- S6: Silent truncation on malformed XML

Will file these as separate issues after merge.

pchuri:
External review (different agent) caught three behavioral regressions vs `main` that the existing test suite missed. Pushed fixes in [8f41c2a](https://github.com/pchuri/confluence-cli/commit/8f41c2a).

**P1 — HTML named entities leaked verbatim**
- `htmlparser2` in xmlMode only decodes the five XML entities (`&amp;`, `&lt;`, `&gt;`, `&quot;`, `&apos;`). HTML named entities (`&nbsp;`, `&eacute;`, `&ndash;`, …) survived as literal text into the markdown output.
- **Fix**: pipe text nodes through `entities.decodeHTML`, then normalize the small subset the original `htmlToMarkdown` explicitly mapped to ASCII (nbsp → space, smart quotes → `"`/`'`, ellipsis → `...`) so user-visible output stays bit-identical for previously-handled chars.

**P2 — `ac:link` without body emitted empty markers**
- `<ac:link ac:anchor="x"><ri:page .../></ac:link>` rendered as `[](#x)`; `<ac:link><ri:url ri:value="…"/></ac:link>` rendered as `[](…)`. The original catch-all dropped these entirely.
- **Fix**: return `''` from the anchor / ri:url branches when no `ac:plain-text-link-body` is present.

**P3 — Panel/shared-block produced extra `>` blank lines**
- `<ac:structured-macro ac:name="panel">…<p>body</p>…</ac:structured-macro>` produced `> **T**\n>\n>\n> body\n>` instead of `> **T**\n>\n> body`.
- **Fix**: trim `walkNodes(body)` before splitting on `\n` so the surrounding `<p>` newlines do not become bracket `>` lines.

Added `entities ^4.5.0` as an explicit dependency (was transitive via htmlparser2). All three regressions now have targeted regression tests.

**Test status**: 406/406 passing (+7 new), lint clean.

pchuri:
Second-round review (same external agent) caught three follow-on regressions from the first entity-decode fix. Pushed [5df64f7](https://github.com/pchuri/confluence-cli/commit/5df64f7).

**Root cause across all three**: my first fix used `decodeHTML(text) → normalizeTypography(text)`, but `normalizeTypography` operated on Unicode codepoints rather than entity sequences. That made it impossible to distinguish `“` that came from `&ldquo;` (should normalize to `"`) from `“` typed literally by the author (should pass through). It also missed the param/attribute paths entirely.

**Replaced with a single `decodeEntities()` that walks `&…;` references in place:**
- ASCII-mapped subset (`&nbsp;`, `&ldquo;`/`&rdquo;`, `&lsquo;`/`&rsquo;`, `&hellip;`) → ASCII chars.
- All other named entities → flow through `entities.decodeHTML` (e.g. `&eacute;` → `é`).
- Numeric refs (`&#NNN;`, `&#xNNN;`) → codepoints.
- **Literal Unicode codepoints in source pass through untouched** because the regex never matches them.

**Findings closed:**
| Finding | Fix |
|---|---|
| Literal `“…”` mangled to ASCII | `decodeEntities` only matches `&…;`, leaves codepoints alone |
| Macro param entities (`<ac:parameter>caf&eacute;</ac:parameter>`) leaked verbatim | `getTextContent` now pipes the collected text through `decodeEntities` |
| Attribute entities (`ri:content-title`, `ri:filename`) leaked verbatim | Decoded at every text-bound attribute read site |

**New regression tests:**
- Literal Unicode preservation (`<p>“hi” ‘a’ …</p>` → `“hi” ‘a’ …`)
- Numeric character references (`&#65;`, `&#x41;`)
- Parameter text decoding (expand title, anchor id)
- Attribute value decoding (`ri:content-title`, `ri:filename`)

**Test status**: 410/410 passing (+4 new), lint clean.

pchuri:
Third-round review surfaced the remaining entity-decode parity gaps. Pushed [768ef53](https://github.com/pchuri/confluence-cli/commit/768ef53).

**Findings closed:**
| # | Path | Fix |
|---|---|---|
| P1 | `getRawText()` (CDATA bodies) | `decodeEntities` at the public-method boundary; internal `_collectRawText` helper for recursion to avoid double-decode |
| P2 | `ac:anchor` URL fragment | `decodeEntities` at read site |
| P2 | `ri:value` URL | `decodeEntities` at read site |
| P3 | generic `<a href>` URL | `decodeEntities` at read site |

Also decoded `ri:space-key` and `<time datetime>` defensively — they rarely carry named entities but consistency removes the "which attribute paths decode and which do not" cognitive load.

**Pattern observation across all three review rounds**: the first PR aimed at structural correctness (the parser), and the three follow-ups all hit the same shape — **path coverage of entity decoding**. The original `htmlToMarkdown` ran a single decode pass at the very end of the pipeline, which silently covered every path: text nodes, attributes, CDATA, params. Replacing the regex chain with a tree walker forced each path to opt in explicitly, and three rounds were needed to find them all.

A `byte-equal-to-main` round-trip test would have caught all of these in CI on PR open. Filing as a follow-up.

**Test status**: 413/413 passing (+3 new), lint clean.

PR #160, opened 2026-05-01
fix: pass attachment ID instead of downloadLink to download()

s2p2-agent:
## Summary

Fixes #27

The bulk download path in `handleAttachmentsList` passed `att.downloadLink` (a relative URL like `/download/attachments/...?version=1&...`) to `client.download()`, which expects an attachment **ID**, **title**, or full **HTTP URL**.

Since the relative URL matched neither ID nor title, `client.download()` threw `"Attachment not found"`.

## Fix

**Line 90** in `src/commands/attachments.ts`:
```diff
-      await client.download(pageId, att.downloadLink, targetPath)
+      await client.download(pageId, att.id, targetPath)
```

The `client.download()\) method already handles resolving the download URL from the attachment ID internally via `this.list()` lookup.

## Tests Added

7 new tests in `tests/client/attachments.test.ts`:
- `normalizeAttachment`: builds full URL from relative path, keeps absolute URLs, falls back to `extensions.downloadLink`
- `download`: resolves by ID, resolves by title, throws on relative URL (bug reproduction), throws on non-existent ID

## Verification

```
bun run lint        # ✅ clean
bun run typecheck   # ✅ clean
bun test            # ✅ 82 pass, 0 fail
bun run build       # ✅ success
```


s2p2-agent:
Wrong repo — moving to S2P2/confluence-cli

PR #163, opened 2026-05-02
feat(client): add limit configurable in getSpaces and spaces command

### LINKED ISSUE #166: [FEATURE] Allow configurable result limit for confluence spaces command
Hi, thanks for this awesome project!

---

**Is your feature request related to a problem? Please describe.**
The `confluence spaces` command returns a fixed number of results (e.g. top 500), and this limit is not configurable.

https://github.com/pchuri/confluence-cli/blob/4185b4a46313ea681a85e25c694ff2ce1fa3fa5f/lib/confluence-client.js#L439-L447

In real-world environments with a large number of spaces, this may be confusing for users because spaces outside of the result set are not shown at all. As a result, users may think a space does not exist, even though it actually exists in Confluence.

Additionally, running the same command multiple times does not change the result, since the same fixed subset is always returned.

https://github.com/pchuri/confluence-cli/blob/179fe4fed4c77bf89589855d69abe8353a0dffe2/bin/confluence.js#L140-L142

Also, the current command description (`List all Confluence spaces`) may be misleading, since the command does not actually list all spaces due to this limitation.

---

**Describe the solution you'd like**
Allow users to specify the number of results via an option, for example:

```text id="q4p2zd"
confluence spaces --limit 1000
```

* Keep the default behavior (limit of 500) unchanged  if not specified

---

**Describe alternatives you've considered**
If making the limit configurable is not desirable, clearly documenting the current limit and sorting behavior would still help users understand why some spaces are missing.

Additionally, updating the command description to reflect the current behavior (e.g. results may be limited) would help avoid confusion.

Alternatively, a flag such as `--all` could also be considered to explicitly retrieve all results.

---

**Use case**
In environments with many spaces, users may need to list and search across a broader set of spaces.

Without this feature, users may incorrectly assume that certain spaces do not exist, which can lead to confusion and inefficiency.

---

**Implementation suggestions**
Introduce a `--limit` option for the `confluence spaces` command.
I've summarized this in the following PR:

- https://github.com/pchuri/confluence-cli/pull/163

Happy to adjust the implementation based on your feedback.

rsk-ymst:
### Description
Previously the spaces API call had a hardcoded limit of 500. 
This change allows callers to request more than 500 spaces by passing an explicit limit to getSpaces(), 
and exposes it as a --limit option on the `confluence spaces` CLI command.

- getSpaces(limit = 500): limit is now a parameter (default 500)
- `confluence spaces --limit <n>`: new CLI option, defaults to 500
- Add tests verifying the limit param is forwarded to the API request

### Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing
- [x]  I have made corresponding changes to the documentation

### Checklist

closes #166

### DIFF
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -139,13 +139,14 @@ program
 // List spaces command
 program
   .command('spaces')
-  .description('List all Confluence spaces')
-  .action(async () => {
+  .description('List Confluence spaces')
+  .option('-l, --limit <limit>', 'Limit number of results', '500')
+  .action(async (options) => {
     const analytics = new Analytics();
     try {
       const config = getConfig(getProfileName());
       const client = new ConfluenceClient(config);
-      const spaces = await client.getSpaces();
+      const spaces = await client.getSpaces(parseInt(options.limit));
       
       console.log(chalk.blue('Available spaces:'));
       spaces.forEach(space => {
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -437,12 +437,12 @@ class ConfluenceClient {
   }
 
   /**
-   * Get all spaces
+   * Get spaces
    */
-  async getSpaces() {
+  async getSpaces(limit = 500) {
     const response = await this.client.get('/space', {
       params: {
-        limit: 500
+        limit
       }
     });
 
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@ A powerful command-line interface for Atlassian Confluence that allows you to re
 - 📖 **Read pages** - Get page content in text or HTML format
 - 🔍 **Search** - Find pages using Confluence's powerful search
 - ℹ️ **Page info** - Get detailed information about pages
-- 🏠 **List spaces** - View all available Confluence spaces
+- 🏠 **List spaces** - View available Confluence spaces
 - ✏️ **Create pages** - Create new pages with support for Markdown, HTML, or Storage format
 - 📝 **Update pages** - Update existing page content and titles
 - 🗑️ **Delete pages** - Delete (or move to trash) pages by ID or URL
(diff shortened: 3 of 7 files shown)


lib/confluence-client.js
```diff
@@ -437,12 +437,12 @@ class ConfluenceClient {
   }
 
   /**
-   * Get all spaces
```
rsk-ymst, line 440: Removed `"all"` since the command does not retrieve all spaces by default.

bin/confluence.js
```diff
@@ -139,13 +139,14 @@ program
 // List spaces command
 program
   .command('spaces')
-  .description('List all Confluence spaces')
```
rsk-ymst, line 142: Removed `"all"` since the command does not retrieve all spaces by default.

pchuri (approved):
Thanks for the contribution! Clean, focused change that addresses #166 well, and the second commit polishing the "all" wording across docs/examples/SKILL.md after self-review is much appreciated.

LGTM — approving. A couple of optional nits that could be addressed here or in a follow-up:

1. **Input validation** — `parseInt(options.limit)` in `bin/confluence.js` will produce `NaN` for `--limit abc` and accepts non-positive values. The `attachments`, `properties`, and `comments` commands already use this pattern for consistency:
   ```js
   const limit = parseInt(options.limit, 10);
   if (Number.isNaN(limit) || limit <= 0) {
     throw new Error('Limit must be a positive number.');
   }
   ```

2. **`parseInt` radix** — minor, but newer commands in this file pass `10` explicitly (`parseInt(x, 10)`).

3. **(Out of scope, just flagging)** The Confluence `/space` endpoint may cap results server-side, so `--limit 1000` may not always return 1000 entries in a single call. True "fetch more than the server cap" support would need pagination via `start` (similar to `getAllProperties`). Worth a follow-up issue if users hit this.

None of these are blockers — the PR is a clear improvement as-is. Thanks again!

rsk-ymst:
Thanks for the review and feedback! I really appreciate it :)
I'll keep an eye on this project going forward !

PR #165, opened 2026-05-03
feat: add --type flag for folder support

### LINKED ISSUE #164: [FEATURE] Support creating Confluence folders
## Summary

The CLI currently hardcodes `type: 'page'` for all content creation operations (`create`, `create-child`). Confluence supports a distinct `folder` content type for organizing pages, but there's no way to create one via the CLI.

## Use Case

When building out a Confluence space programmatically (e.g., from a CI/CD pipeline or AI agent), it's useful to create folders as organizational containers rather than pages. Folders behave differently from pages in the Confluence UI — they appear with a folder icon and are intended for grouping, not content.

## Proposed Solution

Add a `--type` flag (defaulting to `page`) to the `create` and `create-child` commands, allowing users to specify `folder` as an alternative content type:

```sh
confluence create "My Folder" MYSPACE --type folder
confluence create-child "Sub Folder" 123456789 --type folder
```

This would require changing the hardcoded `type: 'page'` in `lib/confluence-client.js` (lines ~1221 and ~1252) to use the flag value instead.

## Alternatives Considered

- Using the Confluence REST API directly, bypassing the CLI
- Creating regular pages as pseudo-folders (loses the semantic distinction and UI treatment)

jasonnissen:
## Summary

- Adds `--type <page|folder>` option to `create` and `create-child` commands (defaults to `page`)
- When `type` is `folder`, body content is omitted from the API payload and `--content`/`--file` become optional
- Includes input validation, dynamic success messages, and unit tests

Closes #164

## Test plan

- [x] All 589 existing tests pass
- [x] 3 new unit tests for folder type handling
- [x] Manually verified: created a folder in a live Confluence Cloud space via `confluence create "Test Folder" FROPS --type folder` — confirmed `/folder/` URL path in response

### DIFF
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -22,6 +22,20 @@ function assertNonEmpty(value, label) {
   }
 }
 
+const VALID_TYPES = ['page', 'folder'];
+
+function assertValidType(type) {
+  if (!VALID_TYPES.includes(type)) {
+    throw new Error(`Invalid type "${type}". Valid: ${VALID_TYPES.join(', ')}`);
+  }
+}
+
+function assertNoBodyForFolder(type, options) {
+  if (type === 'folder' && (options.file || options.content)) {
+    throw new Error('--file/--content is not allowed with --type folder (folders have no body).');
+  }
+}
+
 function handleCommandError(analytics, commandName, error) {
   analytics.track(commandName, false);
   console.error(chalk.red('Error:'), error.message);
@@ -228,11 +242,8 @@ program
     try {
       assertNonEmpty(title, 'title');
       assertNonEmpty(spaceKey, 'spaceKey');
-
-      const VALID_TYPES = ['page', 'folder'];
-      if (!VALID_TYPES.includes(options.type)) {
-        throw new Error(`Invalid type "${options.type}". Valid: ${VALID_TYPES.join(', ')}`);
-      }
+      assertValidType(options.type);
+      assertNoBodyForFolder(options.type, options);
 
       const config = getConfig(getProfileName());
       assertWritable(config);
@@ -280,11 +291,8 @@ program
     try {
       assertNonEmpty(title, 'title');
       assertNonEmpty(parentId, 'parentId');
-
-      const VALID_TYPES = ['page', 'folder'];
-      if (!VALID_TYPES.includes(options.type)) {
-        throw new Error(`Invalid type "${options.type}". Valid: ${VALID_TYPES.join(', ')}`);
-      }
+      assertValidType(options.type);
+      assertNoBodyForFolder(options.type, options);
 
       const config = getConfig(getProfileName());
       assertWritable(config);
@@ -2097,6 +2105,8 @@ module.exports = {
     sanitizeFilename,
     assertWritable,
     assertNonEmpty,
+    assertValidType,
+    assertNoBodyForFolder,
     handleCommandError,
   },
 };
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -218,15 +218,21 @@ program
 // Create command
 program
   .command('create <title> <spaceKey>')
-  .description('Create a new Confluence page')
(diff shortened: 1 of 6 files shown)


bin/confluence.js
```diff
@@ -218,16 +218,22 @@ program
...
+  .option('--type <type>', 'Content type (page, folder)', 'page')
   .action(async (title, spaceKey, options) => {
     const analytics = new Analytics();
     try {
       assertNonEmpty(title, 'title');
       assertNonEmpty(spaceKey, 'spaceKey');
 
+      const VALID_TYPES = ['page', 'folder'];
+      if (!VALID_TYPES.includes(options.type)) {
+        throw new Error(`Invalid type "${options.type}". Valid: ${VALID_TYPES.join(', ')}`);
```
pchuri: Minor / non-blocking: this `VALID_TYPES` block is duplicated in `create-child` at line 284. Consider hoisting it to a single module-level constant (and maybe a small `assertValidType(type)` helper) so the two stay in sync if we ever add a third type.

bin/confluence.js
```diff
@@ -242,13 +248,14 @@ program
         content = fs.readFileSync(options.file, 'utf8');
       } else if (options.content) {
         content = options.content;
-      } else {
+      } else if (options.type !== 'folder') {
```
pchuri, line 262: Non-blocking observation: if the user passes `--type folder` together with `--content` or `--file`, the content is read (and the file's existence is even validated above) but then silently dropped at the client layer because the folder payload omits `body`. Someone might reasonably expect their folder to have been created with content. Worth either erroring out, or printing a small notice like `Note: --content is ignored for --type=folder` so it doesn't turn into a confused debugging session. Same applies to `create-child` around line 307.

pchuri (commented):
Thanks for the PR — nicely scoped change. The implementation is clean, all 589 tests pass on my end, lint is clean, and the live Confluence Cloud verification gives good confidence that `type: 'folder'` works against the v1 endpoint. Internal callers in `copy-tree` correctly stay on the default `page`, so there's no unintended fallout.

Two documentation updates would be worth folding into this PR before merging:

1. **README.md** — the commands table around lines 697-698 lists the `create` / `create-child` options but doesn't include `--type`. A row plus a short example in the usage section would help users discover the flag.

2. **plugins/confluence/skills/confluence/SKILL.md** — the `create` (lines ~256-279) and `create-child` (lines ~281-293) sections don't mention `--type` / folder. Since one of the main use cases this targets is AI agents structuring spaces programmatically, the SKILL doc is exactly where those agents will look — keeping it in sync seems important for this feature in particular.

Left a couple of small inline notes as well, both non-blocking. Thanks again for the contribution!

bin/confluence.js
jasonnissen: Done in 92331a4. Hoisted to a module-level `VALID_TYPES = ['page', 'folder']` plus an `assertValidType(type)` helper near the other input assertions in `bin/confluence.js`. Both `create` and `create-child` now just call `assertValidType(options.type)`, and the helper is exported via `_test` so it has unit coverage in `tests/cli-entry.test.js`.

bin/confluence.js
jasonnissen: Done in 92331a4. Went with the error-out option — added an `assertNoBodyForFolder(type, options)` helper and called it in both `create` and `create-child` right after type validation. Combining `--type folder` with `--file` or `--content` now throws:

> `--file/--content is not allowed with --type folder (folders have no body).`

Felt more consistent with the existing input-validation pattern (`assertNonEmpty`, "Either --file or --content option is required") than a warning that the user might miss. Covered by unit tests in `tests/cli-entry.test.js`. Also documented the rule in the SKILL.md and README updates.

pchuri (approved):
LGTM — all review feedback addressed cleanly. Nice call making the body+folder combination a hard error rather than a warning. Thanks for the quick turnaround!

PR #170, opened 2026-05-04
fix(walker): preserve attributes on <ul>/<ol>/<li> in markdown→storage

devlimits:
### Description

Addresses follow-up 1 from the #153 review (`<ul>` / `<ol>` missing `renderAttrs`).

`<ul>`, `<ol>`, and `<li>` were the only block-level elements in `lib/html-to-storage.js` that dropped their attributes. The `<table>` family already threaded `renderAttrs(node.attribs)` through, so the same plumbing is added here. For raw-HTML input via `--input-format html`, this means `<ul class="task-list">` and `<ol start="3">` now reach the storage payload with attributes intact instead of as bare `<ul>` / `<ol>`.

The `<ul>` / `<ol>` and `<table>` cases are kept as separate `case` blocks (rather than collapsed into one) to preserve the per-domain whitelist intent the rest of `dispatchTag` follows.

For `<li>`, the opening tag is hoisted into an `open` variable so both wrap and unwrap branches share it (same pattern as `<th>` / `<td>`).

### Verification (Confluence Cloud)

| Attribute | Storage save | Notes |
|-----------|--------------|-------|
| `class` on `<ul>` / `<ol>` / `<li>` | preserved | |
| `style` on `<ul>` / `<ol>` / `<li>` | preserved (normalized) | |
| `start` on `<ol>` | preserved | |
| `id`, `data-*`, `target`, `rel`, `type`, `value`, `reversed` | stripped server-side | Safe to emit anyway |

Also verified that `class` position on `<li>` vs the inner `<p>` does not affect storage preservation or rendered HTML — the current walker output shape (`<li class="x"><p>content</p></li>`) is fully accepted.

### Type of Change

- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

### Testing

- [x] Tests pass locally with my changes (633 passing, was 630)

4 new regression tests in `tests/html-to-storage.test.js`:
- `<ul>` with `class` + `style`
- `<ol>` with `class` + `start`
- `<li>` with `class` on the wrap path
- `<li>` with attribute on the unwrap path (block child suppresses `<p>` wrap)

### Out of scope

Round-trip preservation. The storage walker strips list attributes on storage → markdown export, so `<ul class>` round-trip loses its class regardless. That side needs its own design call (markdown has no native attribute syntax).

### DIFF
--- a/lib/html-to-storage.js
+++ b/lib/html-to-storage.js
@@ -344,10 +344,11 @@ function dispatchTag(node, ctx) {
     return `<img${renderAttrs(node.attribs)}>`;
   case 'ul':
   case 'ol':
-    return `<${node.name}>${walkChildren(node, ctx)}</${node.name}>`;
+    return `<${node.name}${renderAttrs(node.attribs)}>${walkChildren(node, ctx)}</${node.name}>`;
   case 'li': {
     const inner = walkChildren(node, ctx);
-    return shouldWrapInP(node) ? `<li><p>${inner}</p></li>` : `<li>${inner}</li>`;
+    const open = `<li${renderAttrs(node.attribs)}>`;
+    return shouldWrapInP(node) ? `${open}<p>${inner}</p></li>` : `${open}${inner}</li>`;
   }
   case 'pre':
     return convertCodeBlock(node, ctx);
--- a/tests/html-to-storage.test.js
+++ b/tests/html-to-storage.test.js
@@ -92,6 +92,26 @@ describe('htmlToStorage', () => {
       expect(htmlToStorage('<ol><li>one</li></ol>'))
         .toBe('<ol><li><p>one</p></li></ol>');
     });
+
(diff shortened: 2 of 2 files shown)


pchuri (approved):
Thanks for the clean follow-up to #153!

The fix mirrors the existing `<th>`/`<td>` `open`-variable pattern exactly, so the change reads as a natural continuation of `dispatchTag` rather than a new shape. Keeping `<ul>`/`<ol>` as separate `case` blocks (instead of folding them into the `<table>` family) is the right call — preserves the per-domain whitelist intent.

Verification table in the description and the four regression tests (wrap path, unwrap path, `<ol start>`, `<li>` with block child) cover the meaningful axes. No new attack surface since `renderAttrs` already runs through `escapeAttrValue` for double-quote handling.

Round-trip preservation being out of scope is the correct framing — that's a separate design call on the storage→markdown side.

LGTM.

PR #188, opened 2026-05-18
Add start offset pagination to search command

barrylv123:
## Summary

Adds a `--start` option to the `search` command so users can request later pages of Confluence search results.

## Changes

- Pass `start` through to the Confluence REST search API.
- Validate `--start` as a non-negative number at the CLI layer.
- Keep displayed result numbering aligned with the requested offset.
- Document the new option in README.
- Add coverage for the client request parameter and CLI help output.

## Testing

- `npx jest --runInBand --no-watchman tests/cli-entry.test.js tests/confluence-client.test.js`
- `npm run lint -- bin/confluence.js lib/confluence-client.js tests/cli-entry.test.js tests/confluence-client.test.js`

### DIFF
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -166,9 +166,15 @@ program
   .command('search <query>')
   .description('Search for Confluence pages')
   .option('-l, --limit <limit>', 'Limit number of results', '10')
+  .option('--start <start>', 'Start index for results', '0')
   .option('--cql', 'Pass query as raw CQL instead of text search')
   .action(withClient('search', async ({ client, analytics }, query, options) => {
-    const results = await client.search(query, parseInt(options.limit), options.cql);
+    const start = parseInt(options.start, 10);
+    if (Number.isNaN(start) || start < 0) {
+      throw new Error('Start must be a non-negative number.');
+    }
+
+    const results = await client.search(query, parseInt(options.limit), options.cql, start);
 
     if (results.length === 0) {
       console.log(chalk.yellow('No results found.'));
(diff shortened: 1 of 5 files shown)


pchuri (commented):
Thanks for picking this up — `--start` has been a missing piece for `search`, and the change is pleasantly small. I appreciate that you offset the displayed numbering (`${start + index + 1}`), updated the README, and added coverage for both the client param and the CLI help.

One required change inline before this can land. The rest of my notes (test coverage for negative `--start`, using the existing `parsePositiveInt` helper in the client, showing `Showing N–M of TOTAL`) are nice-to-haves and can be follow-ups.

bin/confluence.js
```diff
@@ -166,9 +166,20 @@ program
   .command('search <query>')
   .description('Search for Confluence pages')
   .option('-l, --limit <limit>', 'Limit number of results', '10')
+  .option('--start <start>', 'Start index for results', '0')
   .option('--cql', 'Pass query as raw CQL instead of text search')
   .action(withClient('search', async ({ client, analytics }, query, options) => {
-    const results = await client.search(query, parseInt(options.limit), options.cql);
+    const limit = parseInt(options.limit, 10);
```
pchuri: **Required before merge:** this introduces a behavior change that's outside the stated scope of the PR.

Previously `parseInt(options.limit)` happily passed `NaN` or `0` through to the API; now `search --limit 0` and `search --limit abc` throw. That's probably the right tightening, but it's unrelated to adding `--start`, so please pick one of:

1. Call this out in the PR description and CHANGELOG so users know `--limit <= 0` now errors, **or**
2. Drop the `limit` validation from this PR and submit it separately, leaving the `--start` change minimal.

Either is fine — I just don't want the `limit` change to slip in silently under a PR titled "Add start offset pagination".

pchuri (approved):
Thanks for the quick turnaround — scoping the PR down to just `--start` is exactly what I was after. Diff is clean, tests cover the new param, and the offset numbering is a nice touch.

The follow-up items from the earlier review (negative-value tests for `--start`, using the existing `parsePositiveInt` helper in the client, surfacing `Showing N–M of TOTAL` in the output) are all non-blocking and can land separately if you're interested. LGTM 🚀

PR #189, opened 2026-05-18
feat: add api command for arbitrary authenticated requests

LoganBarnett:
## Summary

- Adds a `confluence api <endpoint>` command modeled after `gh api`, allowing users to make authenticated requests to any Confluence REST endpoint (labels, restrictions, groups, audit, v2 API, etc.) without falling back to `curl`.  It's kind of a release valve for API calls that `confluence-api` hasn't had a chance to support yet.
- Adds `rawRequest` method to `ConfluenceClient` supporting relative, absolute, and full URL endpoints.
- Supports `-X`, `-f`, `-H`, `--input`, `--jq`, `-i`, `--silent` options with read-only profile enforcement.

## Test plan

- [x] 12 unit tests for `rawRequest` (relative/absolute/full URLs, methods, headers, auth, errors).
- [x] 14 integration tests for the `api` command (field/header parsing, read-only enforcement, method auto-detection, `--input`, `--silent`, help output).
- [x] All 197 existing tests still pass.
- [x] Lint clean.

### DIFF
--- a/bin/confluence.js
+++ b/bin/confluence.js
@@ -15,6 +15,8 @@ const registerAttachmentCommands = require('./commands/attachments');
 const registerPropertyCommands = require('./commands/properties');
 const registerCommentCommands = require('./commands/comment');
 const registerExportCommand = require('./commands/export');
+const registerApiCommand = require('./commands/api');
+const { readStdin } = require('../lib/stdin-utils');
 
 function assertWritable(config) {
   if (config.readOnly) {
@@ -51,15 +53,6 @@ function handleCommandError(analytics, commandName, error, onExtra = null) {
     try { onExtra(error); } catch { /* keep error path robust if hint code throws */ }
   }
   process.exit(1);
-}
-
-async function readStdin() {
-  process.stdin.setEncoding('utf8');
-  let data = '';
-  for await (const chunk of process.stdin) {
-    data += chunk;
-  }
-  return data;
 }
 
 // Wraps a command action with the standard analytics + client + error pipeline.
@@ -487,6 +480,8 @@ registerPropertyCommands(program, { withClient });
 registerCommentCommands(program, { withClient });
 
 registerExportCommand(program, { withClient });
+
+registerApiCommand(program, { getProfileName, readStdin });
 
 // Copy page tree command
 program
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -2074,6 +2074,33 @@ class ConfluenceClient {
     }
     return parsed;
   }
+
+  async rawRequest(method, endpoint, options = {}) {
+    const { headers = {}, params, data } = options;
+
+    let url;
+    if (/^https?:\/\//i.test(endpoint)) {
+      url = endpoint;
+    } else if (endpoint.startsWith('/')) {
+      url = `${this.protocol}://${this.domain}${endpoint}`;
+    } else {
+      url = endpoint;
+    }
+
+    const response = await this.client.request({
+      method: method.toUpperCase(),
+      url,
+      headers,
+      params,
+      data,
+    });
+
+    return {
+      status: response.status,
+      headers: response.headers,
+      data: response.data,
+    };
+  }
 }
 
 ConfluenceClient.createLocalConverter = function () {
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@ A powerful command-line interface for Atlassian Confluence that allows you to re
 - 🛠️ **Edit workflow** - Export page content for editing and re-import
 - 🔀 **Profiles** - Manage multiple Confluence instances with named configuration profiles
 - 🔒 **Read-only mode** - Profile-level write protection for safe AI agent usage
+- 🌐 **Raw API requests** - Make arbitrary authenticated requests to any Confluence endpoint (like `gh api`)
 - 🔄 **Format conversion** - Convert between Markdown, HTML, Storage, and text formats locally (no server required)
 - 🔧 **Easy setup** - Simple configuration with environment variables or interactive setup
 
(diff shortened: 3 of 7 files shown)


pchuri (commented):
Thanks a lot for the contribution! 🙌 A `gh api`-style escape hatch is genuinely useful for endpoints we haven't wrapped yet, and the test coverage (12 unit + 14 integration) is impressive — that's the right amount of rigor for something this generic.

I left a few inline comments. The two blocking ones (from my perspective) are (1) the file layout and (2) how absolute paths interact with `apiPath` on Cloud. The rest are nice-to-haves that could be follow-ups if you'd rather keep this PR focused. Either way, very nice work.

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
     }
   });
 
+// API command (arbitrary authenticated requests, modeled after gh api)
+program
```
pchuri: **Architecture nit.** PR #186 split this file into per-domain modules under `bin/commands/`. To stay consistent with that direction, would you mind moving these ~135 lines into a new `bin/commands/api.js` and wiring it up the same way the other domain commands are? Right now this is the only new command added directly to `bin/confluence.js` since the refactor.

lib/confluence-client.js
```diff
@@ -2102,6 +2102,33 @@ class ConfluenceClient {
...
   }
+
+  async rawRequest(method, endpoint, options = {}) {
+    const { headers = {}, params, data } = options;
+
+    let url;
+    if (/^https?:\/\//i.test(endpoint)) {
+      url = endpoint;
+    } else if (endpoint.startsWith('/')) {
+      url = `${this.protocol}://${this.domain}${endpoint}`;
```
pchuri, line 2085: **Cloud-vs-Server path mismatch — probably the most important comment.**

On Confluence Cloud, `apiPath` is typically `/wiki/rest/api`. By treating any `/`-prefixed endpoint as bypassing `apiPath`, the README's marquee example

```bash
confluence api /rest/api/content/123456789/label
```

resolves to `https://{domain}/rest/api/content/...` (404 on Cloud) instead of `https://{domain}/wiki/rest/api/content/...`. So the documented examples only work on Server/DC, while Cloud users have to know to write `/wiki/rest/api/...`.

A couple of options:
1. Document explicitly that absolute paths bypass `apiPath`, and update the examples to use `/wiki/rest/api/...` for Cloud.
2. Introduce a third tier: treat `rest/api/...` (no leading slash) as "relative to apiPath" (which axios already does via baseURL), `/...` as "relative to host but auto-prepend `apiPath` when it looks like a REST endpoint", and `https://...` as absolute.

Option 1 is cheaper and matches the `gh api` mental model. Either is fine — but the current behavior + current README will silently break for Cloud users.

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
...
+        }
+        output += '\n';
+      }
+
+      const bodyStr = typeof result.data === 'string' ? result.data : JSON.stringify(result.data, null, 2);
+
+      if (options.jq) {
+        const { execSync } = require('child_process');
+        try {
+          const filtered = execSync(`jq ${JSON.stringify(options.jq)}`, {
```
pchuri: **Shell quoting / minor injection surface.**

`JSON.stringify` wraps the expression in double quotes, but the shell still interprets `$`, backticks, and `\` inside double quotes. For example `--jq '.[] | select(.name == "$HOME")'` will have `$HOME` expanded by the shell before `jq` ever sees it. The blast radius is the user's own machine, but it's also trivial to sidestep:

```js
const { spawnSync } = require('child_process');
const r = spawnSync('jq', [options.jq], {
  input: typeof result.data === 'string' ? result.data : JSON.stringify(result.data),
  encoding: 'utf-8',
});
```

Args go straight to the binary, no shell involved.

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
...
+            input: typeof result.data === 'string' ? result.data : JSON.stringify(result.data),
+            encoding: 'utf-8',
+          });
+          output += filtered;
+        } catch {
+          process.exit(2);
```
pchuri: When `jq` isn't installed (`ENOENT`) or the expression is invalid, this swallows the cause and exits with code 2 with no message. A one-line `console.error('jq failed:', err.message)` (or specifically detecting ENOENT and saying "jq is not installed") would save users a debugging round-trip.

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
...
+      }
+
+      // Determine method
+      const hasFields = Object.keys(fields).length > 0;
+      let body = undefined;
+      if (options.input) {
+        const fs = require('fs');
+        let raw;
+        if (options.input === '-') {
+          raw = fs.readFileSync(process.stdin.fd, 'utf-8');
```
pchuri: PR #184 (`fix(convert): read stdin as a stream and guard against TTY-only invocations`) hit this exact pattern in the `convert` command — `fs.readFileSync(process.stdin.fd, ...)` can hang on a TTY and has issues with large inputs. Could you reuse the stream-based stdin helper from that fix to stay consistent?

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
...
+program
+  .command('api <endpoint>')
+  .description('Make an authenticated API request (like gh api)')
+  .option('-X, --method <method>', 'HTTP method (default: GET, auto-POST when body provided)')
+  .option('-f, --field <key=value>', 'Add a request field (repeatable)', (v, a) => { a.push(v); return a; }, [])
+  .option('-H, --header <key:value>', 'Add a request header (repeatable)', (v, a) => { a.push(v); return a; }, [])
+  .option('--input <file>', 'Read body from file (use - for stdin)')
+  .option('--jq <expression>', 'Filter response with jq')
+  .option('-i, --include', 'Include response status and headers')
+  .option('--silent', 'Suppress all output')
```
pchuri: Minor: the description says "Suppress all output", but the read-only block and the `catch` handler still write to stderr. Either tighten the implementation (`--silent` ⇒ also swallow stderr on errors) or relax the doc to "Suppress success output." Most CLIs go with the latter — script consumers usually still want to see errors.

bin/confluence.js
```diff
@@ -1952,6 +1952,141 @@ profileCmd
...
+
+      const hasBody = hasFields || body !== undefined;
+      const method = (options.method || (hasBody ? 'POST' : 'GET')).toUpperCase();
+
+      // Read-only enforcement
+      const WRITE_METHODS = ['POST', 'PUT', 'PATCH', 'DELETE'];
+      if (config.readOnly && WRITE_METHODS.includes(method)) {
+        console.error(chalk.red('Error: This profile is in read-only mode. Write operations are not allowed.'));
+        console.error(chalk.yellow('Tip: Use "confluence profile add <name>" without --read-only, or set readOnly to false in config.'));
+        process.exit(1);
```
pchuri: `analytics.track('api', false)` is missing here, and on the field/header parse error paths above (lines 1978, 1989). The bottom `catch` tracks failures, but these early `process.exit(1)` branches don't, so the metric will under-count failures.

LoganBarnett:
Thanks for the thorough review!  I'll take a look at these ASAP :)

pchuri:
Hey @LoganBarnett — thanks again for the PR! Looks like you've been heads-down on other projects (no pressure!). I'd like to get this landed, so I'm going to push the review fixes directly to your branch with you as co-author. Happy to revert if you'd prefer to do it yourself — just let me know!

LoganBarnett:
@pchuri thanks!  Sorry about that - I've been traveling and computer time has been sporadic.  Looks like stuff went in quickly though.  Thanks for your work on this!

PR #190, opened 2026-05-20
feat(auth): add 'none' auth type for reverse-proxy injected credentials

NathanRignall:
## Summary

- Add `none` as a new value for `--auth-type` / `CONFLUENCE_AUTH_TYPE`.
- When `authType === 'none'`, the client builds requests with no `Authorization` or `Cookie` header. Designed for setups where a local reverse proxy injects credentials on the wire (SPNEGO/Kerberos, mTLS-at-edge, header injection).
- `validateAuthConfig` short-circuits for `none`, so token / email / cookie / mTLS fields are not required at any layer (CLI flags, env vars, profile file, interactive prompts).
- `hasEnvAuth` in `getConfig` accepts `CONFLUENCE_AUTH_TYPE=none` alone — `CONFLUENCE_DOMAIN` + `CONFLUENCE_AUTH_TYPE=none` is enough to drive the env path.
- README: add profile-init example, env-var example, and an explanatory paragraph matching the existing `mtls` / `cookie` sections; update the `--auth-type` flag listing.

## Why

Behind a reverse proxy that auto-injects auth headers, users had to invent a placeholder token to satisfy the existing required-credentials checks. A first-class `none` mode makes the intent explicit and removes the workaround.

## Behavior preservation

- All existing auth types (`basic`, `bearer`, `mtls`, `cookie`) are unchanged.
- `none` is only selected explicitly — `normalizeAuthType` still defaults to `basic` / `bearer` based on email presence.
- The 401 interceptor gains a `none`-specific hint pointing at the proxy; other hints are untouched.

## Test plan

- [x] `npm test` — 689/689 pass (+4 new: 2 in `tests/config.test.js`, 2 in `tests/confluence-client.test.js`)
- [x] `npm run lint` — clean
- [x] Manual: `node bin/confluence.js init --auth-type none --domain confluence.example.com --protocol https` succeeds without prompting
- [x] Manual: pointed CLI at a local Node HTTP inspector with `CONFLUENCE_AUTH_TYPE=none` — confirmed no `authorization` / `cookie` on the outgoing request
- [x] Manual: real round-trip through a reverse-proxy setup works

### DIFF
--- a/README.md
+++ b/README.md
@@ -164,6 +164,14 @@ confluence --profile sso init \
   --cookie "JSESSIONID=abc123; XSRF-TOKEN=xyz789"
 ```
 
+**Reverse-proxy / no-auth profile** (credentials injected upstream):
+```bash
+confluence --profile proxy init \
+  --domain "confluence.internal" \
+  --api-path "/rest/api" \
+  --auth-type "none"
+```
+
 **Hybrid mode** (some fields provided, rest via prompts):
 ```bash
 # Domain and token provided, will prompt for auth method and email
@@ -176,7 +184,7 @@ confluence init --email "user@example.com" --token "your-api-token"
 **Available flags:**
 - `-d, --domain <domain>` - Confluence domain (e.g., `company.atlassian.net`)
 - `-p, --api-path <path>` - REST API path (e.g., `/wiki/rest/api`)
-- `-a, --auth-type <type>` - Authentication type: `basic`, `bearer`, `mtls`, or `cookie`
+- `-a, --auth-type <type>` - Authentication type: `basic`, `bearer`, `mtls`, `cookie`, or `none`
 - `-e, --email <email>` - Email or username for basic authentication
 - `-t, --token <token>` - API token or password
 - `-c, --cookie <cookie>` - Cookie for Enterprise SSO authentication (e.g., `"JSESSIONID=..."`)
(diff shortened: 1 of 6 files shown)


pchuri (approved):
Thanks for the PR — really clean, well-scoped change. Approving.

Verified locally: `npm test` passes (689/689) and `npm run lint` is clean on this branch.

### What I liked

- **Explicit by design**: `normalizeAuthType` never infers `none` — it has to be selected on purpose. Good safety property for an auth-bypass mode.
- **Validation short-circuit**: the early return in `validateAuthConfig` (lib/config.js:128-131) is placed *before* the `!auth.token` check, so `none` correctly skips the "Bearer or basic authentication requires a token" path that would otherwise misfire.
- **Header path is airtight**: `buildAuthHeader()` → `null` and `buildAuthHeaders()` → `{}` for `none`, so the reused `buildAuthHeaders()` call inside `downloadAttachment` (lib/confluence-client.js:1056) can't leak credentials in this mode either.
- **Docs are consistent**: the README profile example, env var example, flag listing, and the new explanatory paragraph all match the existing `mtls` / `cookie` sections — easy to discover.

### Non-blocking suggestions (happy to merge as-is)

1. **401 hint ordering** — in the response interceptor (lib/confluence-client.js:74-105), `isScopedToken()` is checked before `authType === 'none'`. In the unlikely topology where a reverse proxy fronts a path containing `/ex/confluence/` or uses an `api.atlassian.com`-like host, users would get the scoped-token hint instead of the new reverse-proxy hint. Moving the `none` branch above `isScopedToken()` would make the hint match the explicit user intent.

2. **Stale credentials when `none` + token/cookie/mtls are both set** — if a user sets `CONFLUENCE_AUTH_TYPE=none` alongside `CONFLUENCE_API_TOKEN`, the token is still copied into the resolved config (lib/config.js:793) but never sent. Not a security issue, but a `console.warn` when `none` is combined with other credentials would help avoid "why isn't my token being used?" confusion.

3. **Test coverage gaps worth considering**:
   - `saveConfig` → `getConfig` round-trip for a stored `{ authType: 'none', domain: ... }` profile (exercises the file-based path, not just env).
   - `validateCliOptions({ authType: 'none' })` returning no errors without email/token/cookie.
   - Non-interactive `initConfig` with `--auth-type none --domain x` taking the `hasRequiredValues` branch added at lib/config.js:644.

Nice work — the motivation in the description is clear and the implementation matches the intent precisely.

PR #197, opened 2026-06-01
fix(config): prevent init crash when existing config lacks a profiles key

### LINKED ISSUE #196: [BUG] init crashes with "Cannot set properties of undefined (setting 'default')" when existing config lacks a profiles key
**Describe the bug**

Running `confluence init` crashes with `❌ Cannot set properties of undefined (setting 'default')` when a config file already exists at `~/.confluence-cli/config.json` in an older/unrecognized **flat** shape that has no top-level `domain` key and no `profiles` key (for example a legacy config using `baseUrl` / `apiToken`).

`readConfigFile()` only migrates a flat config when `raw.domain && !raw.profiles` (lib/config.js). A flat config that uses `baseUrl`/`apiToken` (no `domain`) falls through and is returned as-is, so `saveConfig()` then runs `fileData.profiles[targetProfile] = config` against an `undefined` `profiles` and throws. The crash is caught and reported via `process.exit(1)`, leaving init unusable until the file is deleted by hand.

**To Reproduce**

1. Place a legacy flat config at `~/.confluence-cli/config.json`:
   ```json
   { "baseUrl": "https://example.atlassian.net/wiki", "restApiPath": "/rest/api", "email": "user@example.com", "apiToken": "xxxx" }
   ```
2. Run `confluence init` and complete the prompts.
3. See error: `❌ Cannot set properties of undefined (setting 'default')`

**Expected behavior**

`init` should not crash. It should recover any unrecognized/legacy config into the multi-profile structure and write a valid `profiles.default` entry.

**Environment (please complete the following information):**
 - OS: macOS 26.2
 - Node.js version: 26.0.0
 - confluence-cli version: 2.11.0
 - Confluence version: Cloud

**Additional context**

Workaround: delete `~/.confluence-cli/config.json` (or hand-write the `profiles` structure) before running `init`.

**Error logs**

```
❌ Cannot set properties of undefined (setting 'default')
```

omartuhintvs:
### Description

Fixes #196.

`confluence init` crashed with `Cannot set properties of undefined (setting 'default')` when `~/.confluence-cli/config.json` already existed in an unrecognized/legacy **flat** shape with no `profiles` key (e.g. an old config using `baseUrl`/`apiToken`, which `readConfigFile()` does not migrate because it only handles the `domain`-based flat format).

In that case `readConfigFile()` returns the raw object as-is, and `saveConfig()` then assigns `fileData.profiles[targetProfile] = config` against an `undefined` `profiles`, throwing.

This PR guards `fileData.profiles` in `saveConfig()` before the assignment, so `init` recovers any unrecognized legacy config into the multi-profile structure instead of crashing. Minimal, defensive change — no behavior change for valid `profiles` configs.

### Type of Change
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing

Added `tests/init-legacy-config.test.js`, which drives `initConfig` non-interactively over a pre-existing no-profiles config file. It fails before the fix (init exits 1 via the crash) and passes after. Full suite: **725 passing** across 18 suites; ESLint clean.

### Checklist
- [ ] Any dependent changes have been merged and published in downstream modules

### Additional Context

The reproducing config has neither a top-level `domain` (so the existing flat-format migration is skipped) nor a `profiles` key, which is exactly the gap this guard closes.

### DIFF
--- a/lib/config.js
+++ b/lib/config.js
@@ -353,6 +353,14 @@ const saveConfig = (configData, profileName) => {
   // Read existing config file (or create new structure)
   const fileData = readConfigFile() || { activeProfile: DEFAULT_PROFILE, profiles: {} };
 
+  // Guard against a pre-existing config file in an unrecognized/legacy shape that
+  // has no usable `profiles` object (e.g. an old flat config that readConfigFile
+  // could not migrate). Without this, the assignment below throws
+  // "Cannot set properties of undefined (setting 'default')".
+  if (!fileData.profiles || typeof fileData.profiles !== 'object') {
+    fileData.profiles = {};
+  }
+
   const targetProfile = profileName || fileData.activeProfile || DEFAULT_PROFILE;
   fileData.profiles[targetProfile] = config;
 
--- a/tests/init-legacy-config.test.js
+++ b/tests/init-legacy-config.test.js
@@ -1,0 +1,97 @@
+const fs = require('fs');
+
+// Mock fs so we can simulate the on-disk config file without touching ~/.confluence-cli
+jest.mock('fs', () => {
+  const actual = jest.requireActual('fs');
+  return {
+    ...actual,
+    existsSync: jest.fn((...args) => actual.existsSync(...args)),
+    readFileSync: jest.fn((...args) => actual.readFileSync(...args)),
+    writeFileSync: jest.fn((...args) => actual.writeFileSync(...args)),
+    mkdirSync: jest.fn((...args) => actual.mkdirSync(...args)),
+    chmodSync: jest.fn(),
+  };
+});
+
+const { initConfig, CONFIG_FILE } = require('../lib/config');
+
+// Make CONFIG_FILE reads/writes hit our in-memory data instead of disk.
+function mockConfigFile(data) {
+  fs.existsSync.mockImplementation((filePath) => {
+    if (filePath === CONFIG_FILE) return data !== null;
+    return jest.requireActual('fs').existsSync(filePath);
+  });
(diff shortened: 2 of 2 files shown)


pchuri (approved):
Thanks for the clean fix and the solid regression test — much appreciated! 🙏

The change is well-targeted. A few things I verified:

- **Root cause is correct.** `saveConfig()` was the only profile-touching function missing a `fileData.profiles` guard — `getConfig()`, `listProfiles()`, `setActiveProfile()`, and `deleteProfile()` all already guard with `fileData.profiles && ...`. This PR closes that gap and makes the module consistent.
- **Guard ordering is right.** Checking `!fileData.profiles` before `typeof fileData.profiles !== 'object'` matters: a `profiles: null` config would otherwise slip through (`typeof null === 'object'`) and still crash. Nice catch.
- **Test is a genuine regression test.** It drives `initConfig` non-interactively over a legacy `baseUrl`/`apiToken` config (no `domain`, no `profiles`), and asserts no crash + a valid `profiles.default` is written. Fails before the fix, passes after.

One **non-blocking** note for a possible follow-up: because `fileData` keeps the original raw object, the rewritten file retains the legacy top-level keys (including the old `apiToken`) alongside the new `profiles`. It's functionally harmless since reads only consult `profiles`, but it does leave a stale token on disk after the user re-runs `init`. Resetting to a fresh `{ activeProfile, profiles }` structure when the shape is unrecognized would clean that up. Totally optional and out of scope for this crash fix.

Approving. Thanks again! ✅

PR #200, opened 2026-06-11
feat: add --format auto option for write commands

a652:
## Summary

Adds an opt-in `--format auto` for `create`, `create-child`, `update`, and `comment`:

- Content that already starts with markup (storage XHTML, e.g. `<p>...\</p>` or `<ac:structured-macro ...>`) is sent as-is
- Plain text or Markdown is converted to Confluence storage XHTML

This is especially useful for older Confluence Server/Data Center versions, which reject bare text in `body.storage.value` (verified against Confluence Server 6.12.1 — create/update/comment all fail without conversion, and work correctly with it).

**The default format remains `storage` — existing behavior is unchanged.**

## Other changes

- Consolidated the duplicated format-conversion logic in `createPage`/`createChildPage`/`updatePage`/`createComment` into a single `toStorageContent()` helper
- Invalid `--format` values now fail with a helpful error listing valid options (previously silently treated as `storage`)
- Command errors now print the API response body, which makes server-side validation failures (common on older Server versions) much easier to diagnose

## Testing

- `npm test`: 734 passed, 18 suites
- New unit tests cover auto-detection for create/update/comment, markup passthrough, and invalid-format errors
- Manually verified end-to-end against a self-hosted Confluence Server 6.12.1: create, create-child, update, comment, and content round-trip (storage and markdown output) all work

### DIFF
--- a/README.md
+++ b/README.md
@@ -564,7 +564,13 @@ confluence create "My New Page" SPACEKEY --content "**Hello** World!" --format m
 
 # Create from a file
 confluence create "Documentation" SPACEKEY --file ./content.md --format markdown
+
+# Auto-detect: keeps storage XHTML as-is, converts plain text/Markdown
+confluence create "My New Page" SPACEKEY --content "**Hello** World!" --format auto
 ```
+
+For create, create-child, update, and comment, `--format auto` detects the content type:
+it preserves content that starts with markup such as `<p>...</p>` or `<ac:structured-macro ...>`, and converts plain text or Markdown to Confluence storage XHTML. This is useful for older Confluence Server/Data Center versions that reject bare text in a `body.storage.value`.
 
 ### Create a Child Page
 ```bash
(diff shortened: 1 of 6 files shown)


pchuri (approved):
Thanks for the thoughtful PR! 🙏 The design is clean: consolidating the duplicated format-conversion logic into a single `toStorageContent()` helper, keeping `storage` as the default (so existing behavior is unchanged), and making `auto` opt-in. The motivation — supporting older Confluence Server/Data Center versions (verified on 6.12.1) that reject bare text in `body.storage.value` — is well justified, and the test coverage is solid. I confirmed the affected test suites pass locally (559 passed).

A few minor, non-blocking notes for a possible follow-up:

1. **`looksLikeMarkupContent` false positives** — The `/^\s*</` heuristic treats any content starting with `<` as markup, so Markdown autolinks (`<https://example.com>`) or plain text beginning with `<` would be sent as-is without conversion. This is the documented intent and `auto` is opt-in, so it's fine — just worth a one-line note in the README as a known limitation.

2. **Format validation skipped for folders** — `normalizeWriteFormat` only runs inside `toStorageContent`, which isn't called for `--type folder`. So `--format wiki --type folder` would pass without the helpful error. Harmless (folders have no body), just slightly inconsistent with "invalid format always errors."

3. **Erroring on unknown formats is a behavior change** — Previously an unknown `--format` silently fell back to `storage`; now it throws. It's documented and arguably a bug fix, but worth keeping in mind for release versioning.

None of these block merge. Approving — nice work! ✅

PR #209, opened 2026-07-02
fix(convert): pass through <br> in markdown so table cells can have line breaks

devlimits:
## Summary

- Add `br` to the inline-HTML passthrough allow-list in `MacroConverter`, so `<br>` survives markdown→storage conversion instead of being escaped to `&lt;br&gt;`.
- `<br>`, `<br/>`, and `<br />` all pass through; a `<br>` inside a code span/block stays escaped as literal text.

## Why

`confluence convert --input-format markdown --output-format storage` (and the markdown input path of `create` / `update`) runs markdown-it with `html: false`, so all raw inline HTML is escaped. The converter works around this by stashing an allow-listed set of inline tags (`u`, `sub`, `sup`, `mark`, `details`, `summary`) around the render and restoring them afterward — but `br` was missing from that list.

As a result a table cell could not contain a line break. `| v1 | 가<br>나<br>다 |` produced `<td><p>가&lt;br&gt;나&lt;br&gt;다</p></td>`. This is the practical way to get a line break inside a table cell, since markdown table grammar allows only inline text per cell. With `br` added, the same input now yields `<td><p>가<br>나<br>다</p></td>`.

## Behavior preservation

- **False positives still rejected** — the `(?=[\s/>])` lookahead keeps `<broken>` and autolink-shaped `<br:foo>` out of the allow-list.
- **Real escaping unchanged** — only allow-listed tags are stashed; `<`, `&`, and non-listed tags (e.g. `<script>`) are still escaped.
- **Code regions unaffected** — `<br>` inside inline / fenced / indented code stays escaped as literal text via the existing code-range exclusion.

## Test plan

- [x] New tests in `tests/macro-converter.test.js`: `<br>` in a table cell passes through; `<br/>` / `<br />` variants pass through; `<br>` inside inline code stays escaped.
- [x] Full suite green (692 tests).

## Out of scope

- The reverse direction (storage→markdown) already renders a `<br>` node as `\n`, so round-tripping a `<br>` back out of a table cell hits a separate, pre-existing markdown-grammar limitation. This PR only fixes the forward (authoring) direction.

### DIFF
--- a/lib/macro-converter.js
+++ b/lib/macro-converter.js
@@ -21,7 +21,7 @@ const STASH_DELIM = '\uE000';
 // The body alternation `"[^"]*"|'[^']*'|[^>]` makes the match quote-aware
 // so a literal `>` inside a quoted attribute value (e.g.
 // `<mark title="1>0">`) does not terminate the tag prematurely.
-const PASSTHROUGH_TAG_RE = /<\/?(?:u|sub|sup|mark|details|summary)(?=[\s/>])(?:"[^"]*"|'[^']*'|[^>])*>/gi;
+const PASSTHROUGH_TAG_RE = /<\/?(?:br|u|sub|sup|mark|details|summary)(?=[\s/>])(?:"[^"]*"|'[^']*'|[^>])*>/gi;
 // Block-level HTML elements that should pass through WITHOUT markdown processing of their content. 
 const PASSTHROUGH_BLOCK_RE = /<(svg|div)(?:\s[^>]*)?>[\s\S]*?<\/\1>/gi;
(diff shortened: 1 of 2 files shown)


pchuri:
Thanks for the contribution, @devlimits! This is a clean, well-scoped fix — adding `br` to the inline passthrough allow-list is exactly the right approach, and it addresses the table-cell line-break case that markdown table grammar can't otherwise express. The added tests (table cell, `<br/>` / `<br />` variants, and the inline-code exclusion) cover the behavior well, and CI is green across Node 18/20/22.

We'll approve this and get it merged. Thanks again!

PR #230, opened 2026-09-06
feat(auth): allow combining a session cookie with bearer/basic auth

frd1201:
## Summary

Some corporate reverse-proxy/SSO gateways (e.g. F5 BIG-IP APM) terminate the connection with their own access-policy session cookie and still forward the request to Confluence, which then requires its own application-level credential (a Personal Access Token or Basic auth) on top. Today neither `authType=cookie` (sends only `Cookie`) nor a plain `bearer`/`basic` authType (sends only `Authorization`) covers this combination — one of the two headers is always dropped.

- `buildAuthHeaders()` now attaches the `Cookie` header whenever a cookie is configured, regardless of `authType`, so it can combine with Bearer/Basic/mTLS.
- `saveConfig()` now persists the `cookie` field regardless of `authType` (previously only kept when `authType === 'cookie'`, silently dropping it for any other auth type on `init`/`profile add`).
- CLI validation now rejects an empty `--cookie` regardless of auth type, instead of only when `authType === 'cookie'`.
- This mirrors `getConfig()`, which already read the `cookie` field back unconditionally — the read path was already general, the write path and header-building just hadn't caught up.

No changes to the interactive `init` wizard's prompts/flow — this is additive and only reachable via explicit `--cookie` + `--auth-type <other>` flags (or the equivalent env vars), so existing configs and workflows are unaffected.

## Test plan
- [x] `npm test` — all existing suites pass unchanged (1135 tests)
- [x] `npm run lint` — clean
- [x] Added `tests/cookie-combined-auth.test.js`: full non-interactive `initConfig` → `saveConfig` → `getConfig` round trip for cookie+bearer and cookie+basic, a regression guard for the no-cookie case, and the empty-`--cookie` validation error
- [x] Added header-building unit tests in `tests/confluence-client.test.js` for bearer+cookie and basic+cookie combinations
- [x] Documented the pattern in the README with a worked example

### DIFF
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -266,7 +266,11 @@ class ConfluenceClient {
     if (authHeader) {
       headers.Authorization = authHeader;
     }
-    if (this.authType === 'cookie' && this.cookie) {
+    // A cookie can be sent alongside any other auth type (not just `cookie`
+    // auth), which supports reverse-proxy/SSO gateways (e.g. F5 BIG-IP APM)
+    // that require their own session cookie in addition to an
+    // application-level Bearer/Basic credential.
+    if (this.cookie) {
       headers.Cookie = this.cookie;
     }
     return headers;
--- a/lib/confluence-client.js
+++ b/lib/confluence-client.js
@@ -269,8 +269,9 @@ class ConfluenceClient {
     // A cookie can be sent alongside any other auth type (not just `cookie`
     // auth), which supports reverse-proxy/SSO gateways (e.g. F5 BIG-IP APM)
     // that require their own session cookie in addition to an
-    // application-level Bearer/Basic credential.
-    if (this.cookie) {
+    // application-level Bearer/Basic credential. `none` stays credential-free
+    // (authentication is entirely the reverse proxy's responsibility).
+    if (this.authType !== 'none' && this.cookie) {
       headers.Cookie = this.cookie;
     }
     return headers;
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ confluence init --email "user@example.com" --token "your-api-token"
 - `-a, --auth-type <type>` - Authentication type: `basic`, `bearer`, `mtls`, `cookie`, or `none`
 - `-e, --email <email>` - Email or username for basic authentication
 - `-t, --token <token>` - API token or password
-- `-c, --cookie <cookie>` - Cookie for Enterprise SSO authentication (e.g., `"JSESSIONID=..."`)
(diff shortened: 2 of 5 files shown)


lib/confluence-client.js
```diff
@@ -266,7 +266,11 @@ class ConfluenceClient {
     if (authHeader) {
       headers.Authorization = authHeader;
     }
-    if (this.authType === 'cookie' && this.cookie) {
+    // A cookie can be sent alongside any other auth type (not just `cookie`
+    // auth), which supports reverse-proxy/SSO gateways (e.g. F5 BIG-IP APM)
+    // that require their own session cookie in addition to an
+    // application-level Bearer/Basic credential.
+    if (this.cookie) {
```
pchuri: This now also applies to `authType === 'none'`. The README ("Reverse-proxy injected authentication") and the 401 hint a few lines above both promise that `none` sends **no** `Authorization` or `Cookie` header, but with this change:

```bash
CONFLUENCE_AUTH_TYPE=none CONFLUENCE_COOKIE=MRHSession=abc confluence ...
```

produces `buildAuthHeaders() === { Cookie: 'MRHSession=abc' }` (I verified this locally). `init --auth-type none --cookie ...` also passes `validateCliOptions` and gets persisted by `saveConfig` now.

Could you keep `none` credential-free? Something like:

```js
if (this.authType !== 'none' && this.cookie) {
  headers.Cookie = this.cookie;
}
```

plus rejecting `--cookie` together with `--auth-type none` in `validateCliOptions`, and a small test for the `none` + cookie case. Thanks!

pchuri (changes requested):
Thanks a lot for this PR! The use case is real, the change is nicely scoped, and the round-trip tests plus README section are much appreciated.

I pulled the branch, ran the suite and lint (all green), and traced the callers. There is one thing I'd like fixed before merging — see the inline comment. Everything else looks good to me.

lib/confluence-client.js
frd1201: Fixed in fc9fc86: `buildAuthHeaders()` now excludes `none` from the cookie branch (`if (this.authType !== 'none' && this.cookie)`), mirroring `buildAuthHeader()`'s existing exclusion. `validateCliOptions()` now also rejects `--cookie` combined with `--auth-type none`, so it can't be silently persisted via `saveConfig()` either. Added a test for `buildAuthHeaders()` with `authType: 'none'` + a configured cookie, and a test that `initConfig` rejects the `none` + `--cookie` combination.

pchuri (approved):
Thanks for the quick follow-up! fc9fc86 addresses the `authType=none` concern exactly as suggested, with tests for both the header and the CLI validation path. Pulled the branch again: 1137 tests passing, lint clean. LGTM 🙌

PR #192, opened 2026-05-28
feat: add plantuml macro support to markdown conversion

PunKHS:
### Description

Add PlantUML macro support to markdown conversion. The storage-walker.js now recognizes ac:name="plantuml" structured macros and converts them to fenced code blocks with the plantuml language tag, following the same pattern as the existing Mermaid macro support.

### Type of Change
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [x] Code refactoring

### Testing
- [x] Tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] New and existing unit tests pass locally with my changes

### Checklist
- [x] My code follows the style guidelines of this project
- [x] I have performed a self-review of my own code
- [ ] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding changes to the documentation
- [x] My changes generate no new warnings
- [ ] Any dependent changes have been merged and published in downstream modules

### Additional Context
- Supports [PlantUML for Confluence macro](https://avono-support.atlassian.net/wiki/spaces/PUML/pages/9699367/Macro+plantuml)
- Built for compatibility with [PlantUML](https://plantuml.com/) diagram syntax

PR #232, opened 2026-09-20
feat(plantuml): add plantumlcloud macro format option

PunKHS:
### Description

The CLI could only emit the legacy `plantuml` macro (diagram source as raw text in `<ac:plain-text-body>`). Sites running the PlantUML Diagrams for Confluence app render from the `plantumlcloud` macro instead, whose markup is URI-encoded, raw-deflated and base64-encoded into a `data` parameter — and this CLI dropped those macros entirely when reading pages.

This PR adds a selectable output format, defaulting to today's behaviour:

- `plantuml` (default) — `<ac:structured-macro ac:name="plantuml">` with the source in CDATA.
- `plantumlcloud` — `<ac:structured-macro ac:name="plantumlcloud">` with `toolbar`, `filename` (`plantuml-diagram-<n>.svg`), `data`, `compressed` and `revision` parameters.

Reference: [Programmatically adding PlantUML diagrams](https://stratus-addons.atlassian.net/wiki/spaces/PDFC/pages/1839333377/Programmatically+adding+PlantUML+diagrams)

Default output is unchanged, so this is a non-breaking feature.

### Type of Change
- [x] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring

### Testing

### Checklist
- [x] Any dependent changes have been merged and published in downstream modules — N/A: no new dependency, `zlib` is built in

`confluence convert --input-format markdown --output-format storage --plantuml-format plantumlcloud`:

```xml
<ac:structured-macro ac:name="plantumlcloud"><ac:parameter ac:name="toolbar">bottom</ac:parameter><ac:parameter ac:name="filename">plantuml-diagram-1.svg</ac:parameter><ac:parameter ac:name="data">UzUxKC5JLCopzc1RNXB0zMlMTlU1MtBVNXYFUk75SarGjkCGY2lJRmpeSWZyYklmfh5QICi1sDS1uASoBaQGqAGqA2IALj3FBfl5xalATaomBql5KUA7AQ==</ac:parameter><ac:parameter ac:name="compressed">true</ac:parameter><ac:parameter ac:name="revision">1</ac:parameter></ac:structured-macro>
```

Converting that storage back to Markdown restores the identical ` ```plantuml ` fence, so both shapes round-trip.

### Additional Context

- The app also renders from `<filename>.svg` / `.png` attachments produced by a user-operated PlantUML server; this PR emits the macro only — attachments remain a follow-up (documented in README).
- `deflateRawSync` output is not byte-identical to `pako`'s, so tests assert decode-compatibility rather than golden bytes.
- The reader does not branch on `compressed` (the doc defines it as mandatory `true`): it always attempts inflation and falls back to the raw payload, which also preserves plain markup.
- Change artifacts: `openspec/changes/add-plantumlcloud-macro-format`.


pchuri (approved):
Thanks for another PlantUML contribution, @PunKHS! This rounds out #192/#193 nicely. LGTM 👍

Verified locally:
- Full test suite passes (1178 tests), lint clean
- `decodePlantuml` reproduces the payload from the vendor docs, and the new parity fixture (`13-plantumlcloud.xml`) is picked up by the fixture runner
- Default output unchanged, so existing users are unaffected

Merging as-is. I'll open follow-up issues for a couple of small things I noticed while reviewing, so they don't hold this up:
- `CONFLUENCE_PLANTUML_FORMAT` should also be added to the Pi env allow-list in `lib/pi/command-runner.js` and to the env table in `plugins/confluence/skills/confluence/SKILL.md`
- Optional hardening: share one env normalizer between `convert` and `create`/`update`, and pass `maxOutputLength` to `inflateRawSync`

Feel free to pick any of those up if you're interested, but no pressure at all. Thanks again!

